PR #2, opened 2022-01-18
Soft 584 create python script

JarvisWeng:


libraries/codegen/boards/boards.yaml
```diff
@@ -0,0 +1,15 @@
+# Message Definitions in plaintext are on the wiki at:
+# https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/50003973/CAN+Message+Definitions
+#
+# If you are making changes to this file please update the corresponding entry
+# on the wiki. If you need to add a new message use a reasonable
+# reserved ID. The higher ID the lower the priority. Generally
+# - 0-13: Critical messages (have ACK)
```
mitchellostler, line 7: Do we need any board-specific instruction here instead of the msg instruction?

libraries/codegen/templates/_structs.h.jinja
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
mitchellostler, line 1: I'm a bit confused how this template works. I'm assuming that all the structs will be generated into one .h file, but does this have any variability in it?

mitchellostler (approved):
Looks good, just a suggestion and a question from me

libraries/codegen/boards/boards.yaml
JarvisWeng: Yeah we should. Probably make a new Confluence page or something else. But I'll edit this as CAN keeps developing

libraries/codegen/templates/_structs.h.jinja
JarvisWeng: Oh currently right now, this template does nothing. That's for another ticket and PR. I just made this for now

PR #3, opened 2022-01-22
[Soft 603] FWXV CI

ShiCheng-Lu:


ShiCheng-Lu:
I added a `--define` option to scons, not sure now we want to have that but right now it just plops it down in the build.

.github/workflows/make_dbc_release.yml
```diff
@@ -0,0 +1,75 @@
+
```
ryandancy, line 1: The DBC-releaser script is a) out of date (it's on ubuntu-16.04 which doesn't work anymore on GitHub Actions) and b) won't be useful for a while since we don't have a DBC-generation script yet. So I'd say we can just remove it for now, that way we don't have script-failed-to-run errors.

.github/workflows/main.yml
```diff
@@ -0,0 +1,143 @@
...
+      #       const fs = require('fs');
+      #       const comment = fs.readFileSync(
+      #         './.github/workflows/uncommitted-codegen-changes.md', 'utf8');
+      #       github.issues.createComment({
+      #         issue_number: context.issue.number,
+      #         owner: context.repo.owner,
+      #         repo: context.repo.repo,
+      #         body: comment,
+      #       });
+      #       core.setFailed() // Fail on uncommitted changed
```
ryandancy, line 131: Same with this, I'm not entirely sure if it'll be applicable in the new CAN implementation anyways so we can just remove the post-a-PR-comment-on-uncommitted-changes stuff for now. We can always paste it back over from firmware_xiv if necessary. (This also includes removing the uncommitted-codegen-changes.md file.)

.github/workflows/main.yml
```diff
@@ -0,0 +1,143 @@
...
+          make --version
+          clang --version
+          clang-format --version
+          .venv/bin/pip3 freeze
+
+      # - name: Format and lint
+      #   run: |
+      #     make test_format
+      #     scons lint
+      #     make pylint
```
ryandancy, line 102: Format and lint is coming soon (Daniel Qu has a PR up for it) so I think it's fine to leave this commented for now, if that PR gets merged before this one then we can uncomment it / set CI up to run format/lint. But we should remove the lint.py file since the other PR will add it and we don't want merge conflicts.

.github/workflows/main.yml
```diff
@@ -0,0 +1,143 @@
+name: CI
+
+on:
+  # Run on all pull requests and all pushes to master
+  push:
+    branches: [master]
```
ryandancy, line 6: `master` is renamed to `main` in FWXV

ryandancy (changes requested):

ryandancy (approved):

PR #4, opened 2022-01-23
Soft 596 create can get functions

AdrianJendo:
Created files for auto generating bms_carrier_getters.h


libraries/codegen/boards/boards.yaml
```diff
@@ -12,4 +12,21 @@
     - centre_console
     - bms_carrier
     - pedal
-  
\ No newline at end of file
+  IntTypes:
```
JarvisWeng, line 15: What is the purpose for IntTypes?

libraries/codegen/templates/_getters.h.jinja
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "bms_carrier_structs.h"
```
JarvisWeng, line 3: This shouldn't be hard coded. You should get which `_getters.h` you need to generate from the python script

_getters.h
```diff
@@ -0,0 +1,89 @@
+#pragma once
```
JarvisWeng, line 1: Remove any autogenerated files please

libraries/codegen/generator.py
```diff
@@ -30,10 +39,10 @@ def main():
     env = jinja2.Environment(loader=templateLoader)
 
     for y in options.yaml_file:
-        data = read_yaml(y)    
-    
+        data = read_yaml(y)
+
         if options.template:
-            write_template(env, options.template, data)
+            write_template(env, options.template, y, data)
```
JarvisWeng, line 45: I was thinking something more like this.
```python
if options.template:
    output = './' + get_file_name(options.template, y, data)
    write_template(env, options.template, data)
```
And not put `get_file_name` in `write_template`

libraries/codegen/generator.py
```diff
@@ -7,10 +7,18 @@ def read_yaml(yaml_file):
         data = yaml.load(f, Loader=yaml.FullLoader)
         return data
     
-def write_template(env, template_name, data, dest="./"):
+def write_template(env, template_name, yaml_path, data, dest="./"):
+    # get the name of the yaml file from the filepath
```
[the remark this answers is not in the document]
AdrianJendo: > Also only add the `file_prefix` to the `data` only if the template requires it. (has '_' in the template)

This logic already happens on line 17

JarvisWeng (approved):
Looks good for now. We'll add more to the templates and scripts as we go

ryandancy (approved):

PR #5, opened 2022-01-23
Soft 601 delay port

Bafran:


ShiCheng-Lu:
since FreeRTOS only has delay for ms precision, we might want to remove delay_us and call taskDelay from delay_ms. Most delay calls from our projects are ms anyway and this saves from multiplying and dividing by 1000.

libraries/ms-common/inc/x86/delay.h
```diff
@@ -0,0 +1,23 @@
...
+// during the delay.
+//
+// Max allowable time is UINT32_MAX in microseconds (4294.967295 seconds).
+// If a longer duration is needed, check the condition you are waiting on in a
+// loop and then call this function again if the condition is not met.
+
+#include <stdint.h>
+
+#include "FreeRTOS.h"
+#include "task.h"
```
ShiCheng-Lu, line 14: these include headers should be in the implementation file instead of header file, as they are not referenced at all in the header.

libraries/ms-common/src/x86/delay.c
```diff
@@ -0,0 +1,9 @@
+#include "FreeRTOS.h"
+#include "task.h"
+
+#include "delay.h"
```
ryandancy, line 4: Nitpick: in an implementation file we generally put the corresponding header first on a line on its own, so you'd put `delay.h` first here.

libraries/FreeRTOS/src/tasks.c
```diff
@@ -1342,7 +1342,6 @@ static void prvAddNewTaskToReadyList( TCB_t * pxNewTCB )
 /*-----------------------------------------------------------*/
 
 #if ( INCLUDE_vTaskDelay == 1 )
-
```
ryandancy, line 1345: We probably shouldn't change the FreeRTOS source so we know we haven't made custom modifications if/when we upgrade, so this should be reverted.

libraries/ms-common/src/x86/delay.c
```diff
@@ -0,0 +1,9 @@
...
+
+void delay_ms(uint32_t t) {
+  const TickType_t xDelay = t / portTICK_PERIOD_MS;
```
mitchellostler, line 7: You can use the method pdMS_TO_TICKS() defined by freeRTOS here, because it will be the most optimized. You also probably don't need to create a new variable, and just return  `vTaskDelay(pdMS_TO_TICKS(t))` directly.

libraries/core/config.json
```diff
@@ -1,13 +1,13 @@
...
+      "x86"
+  ],
+  "arm_libs": [
+      "stm32f0xx"
+  ],
+  "mocks": {
+      "mock": [
+          "status_impl_update"
+      ]
+  }
```
ryandancy, line 12: Nitpick: the indentation here. We probably won't have autoformatting for JSON files so I'd prefer this to be consistent, and we use 2 space indents everywhere else, so let's use 2 space indents consistently here.

ryandancy (commented):
Seems everyone is reviewing this PR at once - I'll just jump in with a couple of nitpicks.

libraries/ms-common/inc/x86/delay.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
mitchellostler, line 1: I noticed that this is placed in the x86 Library, but this should be the same for both stm and x86. I think you can move these out to the inc and src folders respectively, unless Ryan has told you differently

ryandancy: Yep this can go in `inc` and `src`, doesn't need to be platform-specific. One of the nice things with FreeRTOS is we have a Linux simulator for it so we can run the same code on x86 and arm/stm32 a lot of the time.

libraries/ms-common/inc/x86/delay.h
```diff
@@ -0,0 +1,11 @@
...
+// Delay function library
+//
```
mitchellostler, line 3: Would also be a good idea to specify that this can only be called in a task, and that calling in main/initialize functions will cause undefined behaviour.

mitchellostler (changes requested):
Looks good, just a few improvements. Let me know if you have any questions

mitchellostler (approved):

PR #6, opened 2022-01-26
SOFT-602: Add scons lint and format targets

danielq987:
Jira: https://uwmidsun.atlassian.net/browse/SOFT-602

## Feature
Ported over linting (`lint.py` for c files, `pylint` for python) and formatting (`clang-format` for c files, `autopep8` for python) to scons.

Commands used are the same as `firmware_xiv`.

## Code Changes
- Add `no-format` and `no-lint` fields to `config.json` to exclude specific libraries from linting/formatting
- Refactored/created a few functions to grab source/header/python files from a directory
- Copied over `lint.py` from `firmware_xiv`

## Notes
- Ran `scons lint` and `scons format` already, but have not been committed yet. Will probably do so after PR is reviewed to reduce clutter?

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -126,8 +126,8 @@ def get_srcs(dir):
 
 # Get header files for a project/library
 def get_inc_files(dir):
-    incs = OBJ_DIR.Dir(str(dir)).Dir('inc').glob('*.h')
-    incs += OBJ_DIR.Dir(str(dir)).Dir('inc').Dir(PLATFORM).glob('*.h')
+    incs = dir.Dir('inc').glob('*.h')
+    incs += dir.Dir('inc').Dir(PLATFORM).glob('*.h')
     return incs
 
 # Get header directories for a project/library
--- a/SConstruct
+++ b/SConstruct
@@ -85,7 +85,6 @@ def parse_config(entry):
         'cflags': [],
         'mocks': {},
         'no_lint': False,
-        'no_format': False
     }
     config_file = entry.File('config.json')
     if not config_file.exists():
@@ -316,12 +315,11 @@ for entry in PROJ_DIRS + LIB_DIRS:
     config = parse_config(entry) 
 
     # Avoid linting/formatting external libraries
-    if not config.get('no_format'):
-        format_files += srcs + incs
-        py_format_files += py_files
     if not config.get('no_lint'):
         lint_files += srcs + incs
         py_lint_files += py_files 
+        format_files += srcs + incs
+        py_format_files += py_files
 
 for entry in PYTHON_DIRS:
     py_files = get_py_files(entry)
--- a/SConstruct
+++ b/SConstruct
@@ -298,7 +298,7 @@ PYTHON_DIRS = []
 
 # Formatter configs
 AUTOPEP8_CONFIG = '-a --max-line-length 100 -r'
-CLANG_FORMAT_CONFIG = '-i style=file'
+CLANG_FORMAT_CONFIG = '-i -style=file'
 
 # Convert a list of paths/Dirs to space-separated paths.
 def dir_list_to_str(dir_list):
--- a/SConstruct
+++ b/SConstruct
@@ -114,36 +114,16 @@ def proj_elf(proj_name):
 def proj_bin(proj_name):
     return proj_elf(proj_name).File(proj_name + '.bin')
 
-# Get all src files for a project/library
-def get_srcs(dir):
-    # Glob the source files from OBJ_DIR because it's a variant dir
-    # See: https://scons.org/doc/1.2.0/HTML/scons-user/x3346.html
-    # str(dir) is e.g. 'projects/example', so this is like build/obj/projects/example/src
-    srcs = OBJ_DIR.Dir(str(dir)).Dir('src').glob('*.[cs]')
-    srcs += OBJ_DIR.Dir(str(dir)).Dir('src').Dir(PLATFORM).glob('*.[cs]')
-    return srcs
-
-# Get header files for a project/library
-def get_inc_files(dir):
-    incs = dir.Dir('inc').glob('*.h')
-    incs += dir.Dir('inc').Dir(PLATFORM).glob('*.h')
-    return incs
-
-# Get header directories for a project/library
-def get_inc_dirs(dir):
-    inc_dirs = [dir.Dir('inc')]
-    inc_dirs += [dir.Dir('inc').Dir(PLATFORM)]
-    return inc_dirs
-
-# Get Python scripts for a project/library
-def get_py_files(dir):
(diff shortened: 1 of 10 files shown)


SConstruct
```diff
@@ -83,6 +84,8 @@ def parse_config(entry):
         'arm_libs': [],
         'cflags': [],
         'mocks': {},
+        'no_lint': False,
```
danielq987, line 87: I'm wondering if we need both a `no_lint` and `no_format` config param. Will there be a case where the two params will be different?

SConstruct
```diff
@@ -112,15 +115,36 @@ def proj_elf(proj_name):
...
+
+# Get header directories for a project/library
+def get_inc_dirs(dir):
+    inc_dirs = [dir.Dir('inc')]
+    inc_dirs += [dir.Dir('inc').Dir(PLATFORM)]
+    return inc_dirs
+
+# Get Python scripts for a project/library
+def get_py_files(dir):
+    return glob.glob('{}/**/*.py'.format(str(dir)), recursive=True)
```
danielq987, line 141: Using Python globbing is a bit out of place here, but I couldn't find a way to glob using Scons recursively.

SConstruct
ryandancy: Good point - I don't think there is a case where we'd want to control format and lint separately. Feel free to combine them into one param.

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+# Linting and Formatting 
+###########################################################
+
+lint_files = []
+format_files = []
+py_lint_files = [] 
+py_format_files = [] 
+
+# Python project directories (e.g /mpxe from firmware_xiv)
+PYTHON_DIRS = []
```
ryandancy, line 298: Leaving `PYTHON_DIRS` empty like this is fine for now, but @vaaranan-y when you add the `py` directory you should also change `PYTHON_DIRS` to be set in the directory setup section of this file like `PROJ_DIRS` and `LIB_DIRS`.

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+
+# Formatter configs
+AUTOPEP8_CONFIG = '-a --max-line-length 100 -r'
+CLANG_FORMAT_CONFIG = '-i style=file'
```
ryandancy, line 302: Copy the .clang-format file from firmware_xiv into the root of fwxv, then clang-format will format the same in the new repo. (Also, I believe it's `-style=file`.)

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+
+    config = parse_config(entry) 
+
+    # Avoid linting/formatting external libraries
+    if not config.get('no_format'):
+        format_files += srcs + incs
+        py_format_files += py_files
+    if not config.get('no_lint'):
+        lint_files += srcs + incs
+        py_lint_files += py_files
```
ryandancy, line 324: We should support only formatting/linting a certain project or library using `--project=X` or `--library=X` - currently we always format/lint everything every time but it would be nice to limit that.

Also, since this is on the top level, it's run every time the SConstruct is read (i.e. on every command, even if it's not a lint or format command) - is it possible to move it to a function you call from `run_lint` and `run_format` so that we're not adding overhead to other commands?

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+    format_cmd = ('echo {} | xargs -r clang-format {}'
+                    .format(dir_list_to_str(format_files), CLANG_FORMAT_CONFIG))
+    subprocess.run(format_cmd, shell=True)
+
+    print('\nFormatting *.py files ...')
+    autopep8_cmd = ('autopep8 {1} -i {0}'
+                    .format(dir_list_to_str(py_format_files), AUTOPEP8_CONFIG))
+    subprocess.run(autopep8_cmd, shell=True)
+
+    print('Done Formatting.')
```
ryandancy, line 351: Instead of using shell tricks with `xargs` to run clang-format/lint/pylint for multiple directories, could you just loop over it in Python? It'll be more robust to different shells that way.

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+
+    print('\nLinting *.py files ...')
+    pylint_cmd = 'echo {} | xargs -r pylint'.format(dir_list_to_str(py_lint_files))
+    subprocess.run(pylint_cmd, shell=True)
+
+def run_format(target, source, env):
+    print('\nFormatting *.[ch] in {}, {} ...'.format(str(PROJ_DIR), str(LIB_DIR)))
+    format_cmd = ('echo {} | xargs -r clang-format {}'
+                    .format(dir_list_to_str(format_files), CLANG_FORMAT_CONFIG))
+    subprocess.run(format_cmd, shell=True)
```
ryandancy, line 344: This seems to print out `No such file or directory` for some reason, maybe it's best to just suppress stdout/stderr for this command?

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+        py_lint_files += py_files 
+
+for entry in PYTHON_DIRS:
+    py_files = get_py_files(entry)
+    py_format_files += py_files
+    py_lint_files += py_files
+
+def run_lint(target, source, env):
+    print('\nLinting *.[ch] in {}, {} ...'.format(PROJ_DIR, LIB_DIR))
+    lint_cmd = 'echo {} | xargs -r python2 ./lint.py'.format(dir_list_to_str(lint_files))
```
ryandancy, line 333: This prints out `Ignoring libraries/stm32f0xx/src/startup_stm32f072.s; not a valid file name (c, cc, h, cpp, cu, cuh)` for me, so we're extraneously including `startup_stm32f072.s` in `lint_files`. Could you filter it to only include `*.[ch]` files?

SConstruct
```diff
@@ -261,6 +285,77 @@ Alias('test', test)
...
+    subprocess.run(lint_cmd, shell=True)
+
+    print('\nLinting *.py files ...')
+    pylint_cmd = 'echo {} | xargs -r pylint'.format(dir_list_to_str(py_lint_files))
```
ryandancy, line 337: In MSXIV we had a list of disabled pylint inspections (in `PYLINT_DISABLE` in the Makefile). We should have a way to disable certain pylint inspections globally in MSXV too. I suggest storing the pylint config in a file called `.pylintrc` and using the `--rcfile` option to load it; that way we don't have to maintain a list of disabled inspections in the middle of the SConstruct.

ryandancy (changes requested):
Great job Daniel!

SConstruct
danielq987: I've merged them, under the name `no_lint`

ryandancy (approved):

PR #7, opened 2022-01-29
SOFT-610: Add 'new' target

danielq987:
JIRA: https://uwmidsun.atlassian.net/browse/SOFT-610

## Feature
Added scons target to make a new project or library from a template.

## Code changes
- Copied over `new_target.py` from `firmware_xiv`. Might need a better place to put this instead of the root directory?


new_target.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
```
danielq987, line 1: in firmware_xiv, this file was in a subdirectory `/make`. should that directory exist in fwxv?

new_target.py
```diff
@@ -0,0 +1,101 @@
...
+This module sets up the folder structure and rules.mk for a new project or library.
+
+Usage: python3 new_target.py project|library name
+"""
+import os
+import argparse
+import textwrap
+from string import Template
+
+RULES_TEMPLATE = Template("""\
```
danielq987, line 13: is `rules.mk` still necessary with the scons buildsystem? im guessing no

SConstruct
```diff
@@ -232,7 +233,7 @@ def get_test_list():
     lib = LIBRARY if LIBRARY else ''
     # Assume only one of project or library is set
     entry = proj + lib
-    if entry:
+    if entry and tests.get(entry):
```
danielq987, line 236: this check is included so that `scons new` doesn't error out... 
although, I'm not sure why the test command is still being called with `scons new`... seems to have something to do with the Depends(...) statement on line 254

new_target.py
ryandancy: You could put it in a `scons` directory, might be a good place for lint.py too.

new_target.py
```diff
@@ -0,0 +1,94 @@
...
+
+    projects/libraries
+    └── name
+        ├── inc
+        ├── README.md
+        ├── rules.mk
+        ├── src
+        └── test
+
+    where rules.mk is required for the project or library to be valid.
```
ryandancy, line 50: Update this documentation since we're no longer using `rules.mk`.

new_target.py
```diff
@@ -0,0 +1,94 @@
...
+
+    proj_path = os.path.join(type_folders[target_type], name)
+    folders = ['src', 'inc', 'test']
+
+    for folder in folders:
+        os.makedirs(os.path.join(proj_path, folder), exist_ok=True)
+
+    deps = DEFAULT_DEPS if target_type == 'project' else []
+
+    config = {"libs": deps}
```
ryandancy, line 72: So it's not buried in the middle of a function, you might want to factor out this into a `generate_config` function since it's important and might get frequently modified.

SConstruct
```diff
@@ -256,6 +257,19 @@ Alias('test', test)
 # Helper targets
 ###########################################################
 
+def make_new_target(target, source, env):
+    # No project or library option provided
+    if not PROJECT and not LIBRARY:
+        raise ValueError("Missing project or library name. Expected --project=... or --library=...")
```
ryandancy, line 263: It would give a nicer error message to just print out this error and then `sys.exit(1)` - raising an exception gives a stack trace that the user shouldn't have to parse through.

SConstruct
ryandancy: Looks like `get_test_list()` is being called regardless of whether the test command is called or not, because it's called in the top-level `Depends` call. This is an ok solution.

ryandancy (changes requested):
Looks great, just a couple nitpicks about documentation and the like.

ryandancy:
Also, could you resolve the merge conflicts from the other PR that just got merged?

new_target.py
```diff
@@ -0,0 +1,104 @@
...
+
+    projects/libraries
+    └── name
+        ├── inc
+        ├── README.md
+        ├── config.json 
+        ├── src
+        └── test
+
+    where rules.mk is required for the project or library to be valid.
```
ryandancy, line 64: Update this line too

ryandancy (approved):

PR #8, opened 2022-01-29
Soft 597 create can set functions

AdrianJendo:
Created generation logic for board setter functions

### DIFF
--- a/libraries/codegen/generator.py
+++ b/libraries/codegen/generator.py
@@ -25,7 +25,8 @@ def write_template(env, template_name, file_path, data):
 
 def process_setter_data(board, data, master_data):
     for message in data["Messages"]:
-        if board in data["Messages"][message].get("target", []):
+        if board in data["Messages"][message].get("target", []) and \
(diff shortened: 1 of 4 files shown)


libraries/codegen/generator.py
```diff
@@ -8,20 +9,46 @@ def read_yaml(yaml_file):
...
+def process_setter_data(board, data, master_data):
+    for message in data["Messages"]:
+        if board in data["Messages"][message].get("target", []) and \
+           data["Messages"][message].get("signals", False) :
+            for signal, length in data["Messages"][message]["signals"].items():
+                master_data["Signals"].append((signal, length['length']))
+
+def parse_board_yaml_files(board):
+    # get the working directory to the boards
+    working_dir = os.getcwd()
```
JarvisWeng, line 35: Yeah, looks good. I was thinking more of adding another option, but we can change that later. Getting cwd inside the script isn't usually the best idea since we might move this file or the board directory later

JarvisWeng (approved):

PR #9, opened 2022-01-29
Add LICENSE

ryandancy:
Just copied over the MIT license from MSXIV (and changed the year to 2022). It's the same as we used for MSXIV so it's probably fine, but if there's a discussion to be had about the license we should use then we can discuss that.


mitchellostler (approved):

mitchellostler:
Do you want to add the death penalty for no readme or should I?

PR #11, opened 2022-02-08
Soft 589 create new can dbcs

AdrianJendo:
Use `python3 libraries/codegen/generator.py -t can_dbcs.dbc.jinja` to generate .dbc file


libraries/codegen/templates/system_can.dbc.jinja
```diff
@@ -0,0 +1,13 @@
+{% set messages = data["Messages"] -%}
```
JarvisWeng, line 1: There's some additional metadata that's required for DBCs to work.
Take a look at the DBC we generate from the old car.
https://github.com/uw-midsun/firmware_xiv/releases/tag/DBC-18-1f6e6c6dfcb077e1583b52d5079fe53dcab2104a

It has some metadata at the top and bottom. 
image
Follow the steps in the https://www.csselectronics.com/pages/can-dbc-file-database-intro to validate the DBC you generate

libraries/codegen/templates/system_can.dbc.jinja
```diff
@@ -0,0 +1,47 @@
...
+  BA_REL_
+  BA_DEF_DEF_REL_
+  BU_SG_REL_
+  BU_EV_REL_
+  BU_BO_REL_
+  SG_MUL_VAL_
+
+BS_:
+
+BU_:
```
JarvisWeng, line 35: I believe `BU_:` is all the boards. So just make this list like so
`BU_: BMS_CARRIER CENTRE_CONSOLE....`

PR #15, opened 2022-02-10
Soft_614 GPIO interrupts

ShiCheng-Lu:


libraries/ms-common/src/x86/gpio_it.c
```diff
@@ -3,54 +3,63 @@
...
 #include <stdlib.h>
 
+#include "FreeRTOS.h"
+#include "event_groups.h"
 #include "gpio.h"
 #include "interrupt_def.h"
 #include "status.h"
 #include "x86_interrupt.h"
 
+#define NOT_REGISTERED 0xFF
```
mitchellostler, line 13: Is this used?

libraries/ms-freertos/inc/FreeRTOSConfig.h
```diff
@@ -92,7 +92,7 @@ extern uint32_t SystemCoreClock;
 #define INCLUDE_xTaskGetIdleTaskHandle 0
 #define INCLUDE_eTaskGetState 0
 #define INCLUDE_xEventGroupSetBitFromISR 1
-#define INCLUDE_xTimerPendFunctionCall 0
+#define INCLUDE_xTimerPendFunctionCall 1
```
mitchellostler, line 95: Do we use this anywhere?

mitchellostler (changes requested):
Looks good, a couple of notes and some questions. Sorry this took so long lol

libraries/ms-common/src/arm/gpio_it.c
```diff
@@ -36,43 +36,35 @@ static uint8_t prv_get_irq_channel(uint8_t pin) {
...
   }
   return STATUS_CODE_UNINITIALIZED;
 }
 
 StatusCode gpio_it_register_interrupt(const GpioAddress *address, const InterruptSettings *settings,
-                                      InterruptEdge edge, GpioItCallback callback, void *context) {
-  if (address->port >= NUM_GPIO_PORTS || address->pin >= GPIO_PINS_PER_PORT) {
+                                      InterruptEdge edge, TaskHandle_t task_to_notify,
+                                      uint8_t bit_to_set) {
+  if (address->port >= NUM_GPIO_PORTS || address->pin >= GPIO_PINS_PER_PORT || bit_to_set > 5) {
```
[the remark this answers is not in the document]
ShiCheng-Lu: I think I tried to bit shift by 5 but missed a >, should compare to 32 ie number of bits of task notify.

libraries/ms-common/src/arm/gpio_it.c
```diff
@@ -36,43 +36,35 @@ static uint8_t prv_get_irq_channel(uint8_t pin) {
...
     return status_code(STATUS_CODE_INVALID_ARGS);
```
mitchellostler, line 50: Add a check to make sure the scheduler has not yet started:
  ```
if (!(xTaskGetSchedulerState() == taskSCHEDULER_NOT_STARTED)) {
    return STATUS_CODE_UNREACHABLE;
  }
```

Comment (mitchellostler) [reply libraries/ms-common/src/arm/gpio_it.c:50]: Additionally, when you use the notification library, you can check the input event against `INVALID_EVENT` (defined as 32)

**Review by JarvisWeng — COMMENTED** [review]
Generally it makes sense. Moving the `edge` to the settings and making `TASK`s into the callbacks.  Only problem I guess if multiple `TASK`s want to catch that GPIO interrupt. But I doubt that's a real concern

File: libraries/ms-common/src/arm/gpio_it.c (hunk 3 above)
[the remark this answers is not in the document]
Comment (ShiCheng-Lu) [reply libraries/ms-common/src/arm/gpio_it.c]: this is super out of date? how are you reviewing this version?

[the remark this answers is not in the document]
Comment (JarvisWeng) [reply libraries/ms-common/src/arm/gpio_it.c]: Oh I believe I had a previous review for you that I never pressed submit. I'll get rid of this

File: libraries/ms-freertos/inc/FreeRTOSConfig.h [hunk 5]
```diff
@@ -91,7 +91,7 @@ extern uint32_t SystemCoreClock;
 #define INCLUDE_uxTaskGetStackHighWaterMark 0
 #define INCLUDE_xTaskGetIdleTaskHandle 0
 #define INCLUDE_eTaskGetState 0
-#define INCLUDE_xEventGroupSetBitFromISR 1
+#define INCLUDE_xEventGroupSetBitFromISR 0
```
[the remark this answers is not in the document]
JarvisWeng: Hm, does it break anything if it's not set to 0? If not then I'm fine with it. We can always set it back later

libraries/ms-common/test/test_gpio_it.c
```diff
@@ -0,0 +1,92 @@
...
+
+void setup_test(void) {}
+
+void teardown_test(void) {}
+
+static uint32_t triggered = 0;
+
+TASK(handler, TASK_MIN_STACK_SIZE) {
+  while (true) {
+    notify_get(&triggered);
```
JarvisWeng, line 47: I think there's needs to be some wrapper for this `notify_get`. Otherwise, users can be using `notify` for multiple things within the `TASK` and they will need to get the `triggered` value. @mitchellostler what do you think?

mitchellostler (approved):
I'll go ahead and move this through.
Sorry for keeping you in suspense ShiCheng

PR #16, opened 2022-02-10
[SOFT-587] Task wrapper library

ryandancy:
This is a rough hash of a wrapper library for making tasks. The purpose of it is to provide a nicer API around the FreeRTOS task creation API and to enforce best practices, like cleaning up with `vTaskDelete(NULL)` at the end of a task.

It's quite rough right now, which mostly means the documentation is lacking. I redid the `leds` project in this style mostly to see how it looks. Would appreciate feedback on the API.


libraries/ms-common/inc/tasks.h
```diff
@@ -0,0 +1,39 @@
...
+  char *name;
+  StackType_t *stack;
+  size_t stack_size;
+  StaticTask_t tcb;
+  void *context;
+} Task;
+
+typedef UBaseType_t TaskPriority;
+
+StatusCode tasks_init_task(Task *task, TaskPriority priority, void *context);
```
JarvisWeng, line 37: Just wondering what `context` would be. Is the context in  `tasks_init_task` has `context` and `Task` different?

libraries/ms-common/src/tasks.c
```diff
@@ -0,0 +1,49 @@
...
+#include <stdbool.h>
+
+#include "FreeRTOS.h"
+#include "log.h"
+#include "status.h"
+
+// Add any setup or teardown that needs to be done for every task here.
+static void prv_task(void *params) {
+  Task *task = params;
+
```
JarvisWeng, line 12: I would also like to add a LOG here. Just like `"<task name> Task is starting!"`

libraries/ms-common/src/tasks.c
```diff
@@ -0,0 +1,49 @@
...
+  // Call the task function. This shouldn't exit.
+  task->task_func(task->context);
+
+  // If the task somehow exits, warn and clean up properly.
+  LOG_WARN("WARNING: Task %s exited.\n", task->name);
+  vTaskDelete(NULL);
```
JarvisWeng, line 18: Yeah we will need a way to clean up semaphores and other data. Maybe add a clean up function pointer to `Task`?

libraries/ms-common/src/tasks.c
```diff
@@ -0,0 +1,49 @@
...
+  if (task->stack_size < TASK_MIN_STACK_SIZE) {
+    LOG_WARN("Task '%s' had too small stack size, defaulting to minimum %d\n", task->name,
+             TASK_MIN_STACK_SIZE);
+    task->stack_size = TASK_MIN_STACK_SIZE;
+  }
+
+  task->context = context;
+  task->handle = xTaskCreateStatic(prv_task, task->name, task->stack_size, task, priority,
+                                   task->stack, &task->tcb);
+  return STATUS_CODE_OK;
```
JarvisWeng, line 44: I would add a `configASSERT` to make sure task is created

projects/leds/src/leds_task.c
```diff
@@ -0,0 +1,38 @@
...
+#include "tasks.h"
+
+static const GpioAddress leds[] = {
+  { .port = GPIO_PORT_B, .pin = 5 },   //
+  { .port = GPIO_PORT_B, .pin = 4 },   //
+  { .port = GPIO_PORT_B, .pin = 3 },   //
+  { .port = GPIO_PORT_A, .pin = 15 },  //
+};
+
+TASK(leds_task, 512) {
```
mitchellostler, line 17: I'm a big fan of this interface, I think it makes things much cleaner

libraries/ms-common/inc/tasks.h
ryandancy: `context` is a generic void pointer that gets passed to the generated task function. Usually I imagine it'll probably be ignored/NULL, but it could be useful in some situations. The `context` field in `Task` is the same, it's just used for convenience in `tasks_init_task` to pass the context to `prv_task`. It doesn't *really* belong there, but the alternative is making a wrapper struct to pass a Task-and-context pair to `prv_task` and then somehow managing that statically.

libraries/ms-common/src/tasks.c
```diff
@@ -0,0 +1,49 @@
+#include "tasks.h"
+
+#include <stdbool.h>
+
+#include "FreeRTOS.h"
+#include "log.h"
+#include "status.h"
+
+// Add any setup or teardown that needs to be done for every task here.
+static void prv_task(void *params) {
```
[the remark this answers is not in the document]
ryandancy: Unfortunately it has to be a void pointer so we can pass it as a task function to `xTaskCreateStatic` :(

libraries/ms-common/src/tasks.c
ryandancy: I think it's probably fine to leave it for now since tasks really shouldn't be executing, and everything is allocated statically. I'm not sure of a clean way to add a cleanup function without complicating the API.

libraries/ms-common/inc/tasks.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
```
mitchellostler, line 1: Do we want a test file for this? I admit it is pretty straight-forward and there won't be any bugs that will pop up, but I think it could be good practice. We could even assign it to a new member, I think that would make a really good first task, to just effectively try to reach every line of coder and get familiar with it.

ryandancy: Hm, I agree. I wasn't quite sure how to test it since it's interacting with the FreeRTOS scheduler, but we'll need a way to test tasks at some point and we should figure it out now.

mitchellostler (approved):

libraries/ms-common/inc/tasks.h
ryandancy: Looks like we'll need to build something on top of Unity to easily support testing whole tasks, that's for a separate ticket.

PR #17, opened 2022-02-10
Soft 603 fwxv ci

ShiCheng-Lu:


ryandancy:
Looks good, did you test to make sure it fails on both format and lint failures?

ShiCheng-Lu:
I don't know why this is failing, llint c files doesn't seem to produce any output. python code could have raised an error somewhere.

libraries/ms-freertos/inc/FreeRTOSConfig.h
```diff
@@ -61,7 +61,7 @@ extern uint32_t SystemCoreClock;
 #define INCLUDE_vTaskDelete 1
 #define INCLUDE_vTaskSuspend 1
 #define INCLUDE_xResumeFromISR 1
-#define INCLUDE_vTaskDelayUntil 1
+#define INCLUDE_xTaskDelayUntil 1
```
mitchellostler, line 64: Do you know why this was changed? I'm pretty sure the API is **v**TaskDelayUntil

ShiCheng-Lu: /* INCLUDE_vTaskDelayUntil was replaced by INCLUDE_xTaskDelayUntil.  Backward
 * compatibility is maintained if only one or the other is defined, but
 * there is a conflict if both are defined. */
 comment from FreeRTOS,
 the function itself is called xTaskDelayUntil which replaced an older vTaskDelayUntil. This Config was old so it still had vTaskDelayUntil but we should be using xTaskDelayUntil

ShiCheng-Lu:
Oh right, since I've changed scons to use cpplint instead of the cpplint.py file renamed to lint.py, this linting via `scons lint` now requires cpplint to be installed via pip. `pip install --upgrade cpplint`
(or you can switch to ubuntu-20.04 box which has everything that's needed installed)

ryandancy (approved):
If we're using cpplint instead of lint.py, should we delete lint.py?

PR #18, opened 2022-02-12
Soft 592 create recv struct

Bafran:


libraries/codegen/bms_carrier_receive_structs.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
JarvisWeng: This was a temporary example file. Please delete if you don't need it any more

libraries/codegen/generator.py
```diff
@@ -28,7 +28,7 @@ def process_setter_data(board, data, master_data):
         if board in data["Messages"][message].get("target", []) and \
            data["Messages"][message].get("signals", False) :
             for signal, length in data["Messages"][message]["signals"].items():
```
JarvisWeng, line 30: Instead of `for signal, length`
Change this to `for signal, signal_data`

Then just add `signal_data["message"] = message` and only append signal and signal_data.

You'll need to rebase this though since another merge will come in and change stuff

libraries/codegen/templates/_receive_structs.h.jinja
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+{% set board = data["Board"] -%}
+{% set signals = data["Signals"] -%}
+
+#include <stdint.h>
+
+typedef struct {
+    {%- for signal, length, message in signals -%}
+        {{length}} {{message}}_{{signal}};
```
JarvisWeng, line 10: This will change slightly to be just the length instead of `uint...` so just make changes accordingly after this PR goes in https://github.com/uw-midsun/fwxv/pull/11/files

libraries/codegen/templates/_setters.h.jinja
```diff
@@ -5,7 +5,7 @@
 
 #include "{{board}}_structs.h"
 
-{% for (signal, length) in signals -%}
+{% for (signal, length, message) in signals -%}
 
 void set_{{board}}_{{signal}}(receive_struct* s, {{length}} val)
```
JarvisWeng, line 10: This should actually be `set_{{message}}_{{signal}}({{board}}_receive_struct.....`

JarvisWeng: and `s->{{board}}_{{signal}} = val;` should be `s->{{message}}_{{signal}} = val;`

PR #20, opened 2022-02-24
Add mutex library

mitchellostler:
Scons is giving me some grief right now for formatting/running, but the library compiles. I'll get to that before we merge.

Something to think about:
- Are we still planning to use status.h in the same way? I figured it was a good idea for now, but we can't exactly do a status_ok_or_return from a task, so may need to improve upon it in the future

(also got rid of the non-unicode characters from new_target)

### DIFF
--- a/libraries/ms-common/inc/mutex.h
+++ b/libraries/ms-common/inc/mutex.h
@@ -9,10 +9,10 @@
 #define BLOCK_INDEFINITELY UINT16_MAX
 
 // Mutex handle is used to access Mutex
-// Mutex Objects must be declared statically 
+// Mutex Objects must be declared statically
 typedef struct Mutex {
   SemaphoreHandle_t handle;
-  StaticSemaphore_t buffer; 
+  StaticSemaphore_t buffer;
 } Mutex;
 
 // Initializes a mutex using provided static entity
--- a/libraries/ms-common/inc/mutex.h
+++ b/libraries/ms-common/inc/mutex.h
@@ -1,0 +1,35 @@
+#pragma once
+// Wrapper library for all mutex and semaphore usage
+
+#include "FreeRTOS.h"
+#include "smphr.h"
+#include "status.h"
+
+#define BLOCK_INDEFINITELY UINT16_MAX
+
+typedef enum MutexCaller {
+  MUTEX_CALLER_TASK = 0,
+  MUTEX_CALLER_ISR,
+}
+
+// Mutex handle is used to access Mutex
+// Mutex Objects must be declared statically 
+typedef struct Mutex {
+  SemaphoreHandle_t handle;
+  StaticSemaphore_t buffer; 
+} Mutex;
+
+// Initializes a mutex using provided static entity
+// Returns STATUS_CODE_OK on success, STATUS_CODE_UNINITIALIZED on failure
+StatusCode mutex_init(Mutex *mutex);
+
+// Locks a Mutex. The locking task is responsible for unlocking the mutex
+// Waits ms_to_wait milliseconds if mutex is already locked, then times out
+// Using BLOCK_INDEFINITELY will cause this method to wait forever on mutex becoming available
+// Returns STATUS_CODE_OK on success, STATUS_CODE_TIMEOUT on timeout
+StatusCode mutex_lock(Mutex *mutex, uint16_t ms_to_wait);
+
+// Unlocks a Mutex. The task which locks the mutex MUST also be the one to unlock it
+// Mutex type specifies whether the method is being called from a task or interrupt handler
+// Returns STATUS_CODE_OK on success, STATUS_CODE_INTERNTAL_ERROR on failure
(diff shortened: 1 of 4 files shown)


JarvisWeng:
> Scons is giving me some grief right now for formatting/running, but the library compiles. I'll get to that before we merge.
> 
> Something to think about:
> 
> * Are we still planning to use status.h in the same way? I figured it was a good idea for now, but we can't exactly do a status_ok_or_return from a task, so may need to improve upon it in the future
> 
> (also got rid of the non-unicode characters from new_target)

Yeah, I used status as well. I think perhaps we should make it like a hard fail? Not sure if we can do that on FreeRTOs but something that kills all the tasks if it fails.

libraries/ms-common/src/mutex.c
```diff
@@ -0,0 +1,42 @@
...
+}
+
+StatusCode mutex_unlock(Mutex *mutex, MutexCaller caller, bool *higher_priority_task_woken) {
+  BaseType_t result, higher_priority_woken = pdFALSE;
+  if (higher_priority_task_woken == NULL) {
+      return STATUS_CODE_INVALID_ARGS;
+  }
+  if(caller == MUTEX_CALLER_ISR) {
+    // Use ISR specific method
+    result = xSemaphoreGiveFromISR(mutex->handle, &higher_priority_woken);
```
JarvisWeng, line 32: Hm, could we also wrap `mutex_unlock` too? So we don't need to pass the caller or bool?

ryandancy:
Are we sure we want to use semaphores? As I recall FreeRTOS mutexes have a nice priority inheritance feature that tries to handle priority inversion, like if a low-priority task is holding a resource but a high-priority task needs it.

mitchellostler:
> Are we sure we want to use semaphores? As I recall FreeRTOS mutexes have a nice priority inheritance feature that tries to handle priority inversion, like if a low-priority task is holding a resource but a high-priority task needs it.

Mutexes are part of the FreeRTOS semaphore. xSemaphoreCreateMutex is the creator, and the object has all the features described, including the priority inversion

libraries/ms-common/src/mutex.c
mitchellostler: Yeah, that's a good idea. I will add those

mitchellostler: Actually I realized this is dumb anyways since mutexes shouldn't be used in interrupts. If we implement a semaphore api I will add this. Will remove the option

libraries/ms-common/inc/mutex.h
```diff
@@ -0,0 +1,37 @@
...
+// Waits ms_to_wait milliseconds if mutex is already locked, then times out
+// Using BLOCK_INDEFINITELY will cause this method to wait forever on mutex becoming available
+// Returns STATUS_CODE_OK on success, STATUS_CODE_TIMEOUT on timeout
+StatusCode mutex_lock(Mutex *mutex, uint16_t ms_to_wait);
+
+// Unlocks a Mutex. The task which locks the mutex MUST also be the one to unlock it
+// Caller specifies whether the method is being called from a task or interrupt handler
+// higher_priority_task_woken informs the caller whether a higher priority task has become unblocked
+// Returns STATUS_CODE_OK on success, STATUS_CODE_INTERNTAL_ERROR on failure
+StatusCode mutex_unlock(Mutex *mutex, MutexCaller caller, bool *higher_priority_task_woken);
```
ryandancy, line 37: It might be a nicer API to have two separate functions, `mutex_unlock` and `mutex_unlock_from_isr`, rather than using a `MutexCaller` parameter. Then we'd only have to pass the `higher_priority_task_woken` parameter in the ISR version since it seems it's always null in the task version.

libraries/ms-common/inc/mutex.h
```diff
@@ -0,0 +1,37 @@
...
+// Locks a Mutex. The locking task is responsible for unlocking the mutex
+// Waits ms_to_wait milliseconds if mutex is already locked, then times out
+// Using BLOCK_INDEFINITELY will cause this method to wait forever on mutex becoming available
+// Returns STATUS_CODE_OK on success, STATUS_CODE_TIMEOUT on timeout
+StatusCode mutex_lock(Mutex *mutex, uint16_t ms_to_wait);
+
+// Unlocks a Mutex. The task which locks the mutex MUST also be the one to unlock it
+// Caller specifies whether the method is being called from a task or interrupt handler
+// higher_priority_task_woken informs the caller whether a higher priority task has become unblocked
+// Returns STATUS_CODE_OK on success, STATUS_CODE_INTERNTAL_ERROR on failure
```
ryandancy, line 36: Typo: `INTERNTAL`

ryandancy (commented):
We'd probably want to test this using the new test API, but I'll have to finish that first...

ryandancy (approved):

ryandancy:
I think it still needs a format but we should be good to merge otherwise

PR #21, opened 2022-02-26
Integrate generator.py into build system

AdrianJendo:
Run the following to test:
`python3 new_target.py project bms_carrier`
then
`scons --project=bms_carrier --platform=x86`

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -78,14 +78,15 @@ LIB_BIN_DIR = BIN_DIR.Dir('libraries')
 
 PLATFORM_DIR = Dir('platform')
 
+CODEGEN_DIR = LIB_DIR.Dir("codegen")
+BOARDS_DIR = CODEGEN_DIR.Dir("boards")
+GENERATOR = CODEGEN_DIR.File("generator.py")
+TEMPLATES_DIR = CODEGEN_DIR.Dir("templates")
+HEADER_OUTPUT_DIR = PROJ_DIR.Dir(PROJECT).Dir("inc")
+LIBRARIES_INC_DIR = LIB_DIR.Dir("ms-common").Dir("inc")
+
 # Put object files in OBJ_DIR so they don't clog the source folders
 VariantDir(OBJ_DIR, Dir('.'), duplicate=0)
-CodgegenDir = LIB_DIR.Dir("codegen")
-BoardsDir = CodgegenDir.Dir("boards")
-Generator = CodgegenDir.File("generator.py")
-TemplatesDir = CodgegenDir.Dir("templates")
-HeaderOutputDir = PROJ_DIR.Dir(PROJECT).Dir("inc")
-LibrariesIncDir = LIB_DIR.Dir("ms-common").Dir("inc")
 
 ###########################################################
 # Targets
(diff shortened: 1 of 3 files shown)


libraries/codegen/generator.py
```diff
@@ -122,22 +119,25 @@ def main():
...
-        write_template(env, options.template, file_path, data)
-    elif options.board: # only for _setters.h
-        data = parse_board_yaml_files(options.board)
-        if options.template:
-            file_path = "./" + options.board + options.template[:-6]
+    if options.template:
+        file_path = get_file_path(
+            options.template,
+            options.board,
+            "./projects/{}/inc/".format(options.board)
```
JarvisWeng, line 126: You shouldn't get the file path from here. The reason why I made the `file_path` in `write_template` generic was so that you could pass the path yourself later.
If you do it like this, you can only use the `generator.py`  with a fixed output path.
You should instead just pass the path you want from the SConstruct.

SConstruct
```diff
@@ -378,6 +378,26 @@ Alias('lint', lint)
 format = Command('format.txt', [], run_format)
 Alias('format', format)
 
+###########################################################
+# Header file generation from jinja templates
+###########################################################
+def generate_header_files():
+    codegen_dir = "libraries/codegen"
+    working_dir = os.getcwd()
+    base_exec = "python3 {}/generator.py -b {}".format(codegen_dir, PROJECT)
```
JarvisWeng, line 387: There should be a check such as if a `project.yaml` doesn't exist then fail

SConstruct
```diff
@@ -378,6 +378,26 @@ Alias('lint', lint)
 format = Command('format.txt', [], run_format)
 Alias('format', format)
 
+###########################################################
+# Header file generation from jinja templates
+###########################################################
+def generate_header_files():
+    codegen_dir = "libraries/codegen"
+    working_dir = os.getcwd()
```
JarvisWeng, line 386: Don't do this in SCons. You should be using Scons objects such as `Dir` or `File` if you want paths or files

SConstruct
```diff
@@ -378,6 +378,26 @@ Alias('lint', lint)
...
+###########################################################
+def generate_header_files():
+    codegen_dir = "libraries/codegen"
+    working_dir = os.getcwd()
+    base_exec = "python3 {}/generator.py -b {}".format(codegen_dir, PROJECT)
+    templates = os.listdir("{}/{}/templates".format(working_dir, codegen_dir))
+
+    for template in templates:
+        if "can_board_ids" in template:
+            env.Execute("{} -y {}/boards/boards.yaml -t {}".format(base_exec, codegen_dir, template))
```
JarvisWeng, line 392: Instead of using `Execute` try making a Scons `Action` or `Builder` or `Method`. This is because we're going to need `targets` from it in order to build a proper dependency tree for Scons to use

SConstruct
```diff
@@ -176,6 +176,43 @@ for entry in PROJ_DIRS + LIB_DIRS:
 # Build all projects when you just run `scons`
 Default([proj.name for proj in PROJ_DIRS])
 
+###########################################################
+# Header file generation from jinja templates
+###########################################################
+def generate_header_files(env, target, source):
```
AdrianJendo, line 182: You were right about this btw, env needs to be the first arg

JarvisWeng: Yeah, thought so

SConstruct
```diff
@@ -176,6 +176,43 @@ for entry in PROJ_DIRS + LIB_DIRS:
...
+    
+    header_files = []
+    for template in templates:
+        if template.name[0] == "_":
+            header_files.append(header_output_dir.File(PROJECT + template.name[:-6]))
+        else:
+            header_files.append(header_output_dir.File(template.name[:-6]))
+
+        if "can_board_ids" in template.name:
+            env.Execute("{} -y {}.yaml -t {}".format(base_exec, boards_dir.File("boards"), template))
```
JarvisWeng, line 198: Oh I forgot to mention. But `can_board_ids.h` is a special case. This actually needs to go into `libraries/ms-common/inc` instead of each project

SConstruct
```diff
@@ -176,6 +176,43 @@ for entry in PROJ_DIRS + LIB_DIRS:
...
+        else:
+            if "_getters" in template.name:
+                env.Execute("{} -y {}.yaml -t {}".format(base_exec, boards_dir.File(PROJECT), template))
+            else:
+                env.Execute("{} -t {}".format(base_exec, template))
+
+    return header_files
+
+codegen_dir = LIB_DIR.Dir("codegen")
```
JarvisWeng, line 207: Make these capitals and move them under the `VariantDir` both. (line 82)

SConstruct
```diff
@@ -176,6 +176,43 @@ for entry in PROJ_DIRS + LIB_DIRS:
...
+boards_dir = codegen_dir.Dir("boards")
+generator = codegen_dir.File("generator.py")
+templates_dir = codegen_dir.Dir("templates")
+header_output_dir = PROJ_DIR.Dir(PROJECT).Dir("inc")
+
+env.AddMethod(generate_header_files, "GenerateHeaderFiles")
+header_targets = env.GenerateHeaderFiles(None, [boards_dir, templates_dir, generator, header_output_dir])
```
JarvisWeng, line 214: Oh yeah your build is failing. This should actually be called above. Probably just before line 147. Or just put it at line 147 and comment it out for now

SConstruct
```diff
@@ -231,10 +268,14 @@ for entry in PROJ_DIRS + LIB_DIRS:
             target = env.Bin(target=output.File(test_file.name + '.bin'), source=target)
 
         # Make test executable depend on the project / library final target
+        proj_elf_target = proj_elf(entry.name)
+        lib_bin_target = lib_bin(entry.name)
+        Depends(proj_elf_target, header_targets)
```
JarvisWeng, line 273: Get rid of these. We will eventually add these just not now. Need to combine everything before I can but them in

SConstruct
```diff
@@ -80,6 +80,12 @@ PLATFORM_DIR = Dir('platform')
 
 # Put object files in OBJ_DIR so they don't clog the source folders
 VariantDir(OBJ_DIR, Dir('.'), duplicate=0)
+CodgegenDir = LIB_DIR.Dir("codegen")
```
JarvisWeng, line 83: Lol, no the variables should be ALL CAPITALS. Like `BUILD_DIR`, `BIN_DIR`, or `OBJ_DIR`.

SConstruct
```diff
@@ -123,6 +129,35 @@ def proj_elf(proj_name):
 def proj_bin(proj_name):
     return proj_elf(proj_name).File(proj_name + '.bin')
 
+###########################################################
+# Header file generation from jinja templates
+###########################################################
+def generate_header_files(env, target, source):
+    BoardsDir = source[0]
```
JarvisWeng, line 136: These are local variables to the function so they don't necessarily need to be capitals. Just make them lower case again

JarvisWeng (approved):

PR #22, opened 2022-02-26
FWXV-58: Reduce build system spam

danielq987:
## Description
With ShiCheng's suggestion, added COMSTR env variables to control output when building projects.

Kind of outside of scope of this PR, but also moved `lint.py` and `new_target.py` to a new `scons` folder and updated their paths accordingly.

## Comments
Running scons using `scons -Q` will help suppress some more scons messages - notably: 
```
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
```


ryandancy:
This is much nicer! I still get some long prints from `ar`, `ranlib`, and `objcopy` - are there more `COMSTR` variables that could make those messages nicer?

danielq987:
> This is much nicer! I still get some long prints from `ar`, `ranlib`, and `objcopy` - are there more `COMSTR` variables that could make those messages nicer?

Unfortunately, I couldn't find something for objcopy, but was able to use ARCOMSTR, RANLIBCOMSTR, and also ASCOMSTR for compiling assembly files.

danielq987:
I couldn't find anything in the Scons Builder documentation for controlling the build output... maybe `objcopy` will just have to stay the way it is. Fortunately the output is generally a lot better than before now.

ryandancy (approved):

PR #24, opened 2022-03-03
[FWXV-11] FreeRTOS queue wrapper

hewittmcg:
Basic wrapper over FreeRTOS queues:

- Adds functions to initialize a static queue and send/receive/peek using a single Queue struct.
- Tested by modifying the queue project from the FreeRTOS workshop to use the queue wrappers and confirming that it still operates the same.

I created this branch before the ticket got moved to the FWXV project, hence the outdated branch name and commit prefixes.


libraries/ms-common/src/queues.c
```diff
@@ -0,0 +1,46 @@
+#include "queues.h"
+#include <string.h>
+
+StatusCode queue_init(Queue *queue) {
+    queue->handle = xQueueCreateStatic(queue->size, queue->item_size, queue->storage_buf, &queue->queue);
```
JarvisWeng, line 5: Hm using this users would have to provide their own storage. Which is fine.
But I would if we can do something like thie
```
StatusCode queue_init(Queue *queue) {
    static buffer[size * item_size] = { 0 };
    ....
}
```
And let the function do it. Probably not very safe though

mitchellostler: We unfortunately can't do this, since this will only allocate the space once, and we need to do it per queue. The other way to go about doing this would be some macro magic, but I think it's fine to just store all the needed parameters in a queue struct, which you can initialize the queue with. This is how it was done with mutex

libraries/ms-common/inc/queues.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+// Basic wrapper for FreeRTOS queues.
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "status.h"
+
+#define QUEUE_DELAY_BLOCKING portMAX_DELAY
+
```
mitchellostler, line 8: Should specify here that this queue struct **must** be declared statically

libraries/ms-common/inc/queues.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+// Basic wrapper for FreeRTOS queues.
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "status.h"
+
+#define QUEUE_DELAY_BLOCKING portMAX_DELAY
+
+typedef struct {
+    uint32_t size;
```
mitchellostler, line 10: Maybe change this to num_queue_items, as this is a little bit vague? Or maybe add a short comment next to each item to say what it is/does.

libraries/ms-common/inc/queues.h
```diff
@@ -0,0 +1,30 @@
...
+typedef struct {
+    uint32_t size;
+    uint32_t item_size;
+    uint8_t *storage_buf;
+    StaticQueue_t queue;
+    QueueHandle_t handle;
+} Queue;
+
+// Create a queue with the parameters specified in settings. Returns STATUS_CODE_OK if successful, STATUS_CODE_INVALID_ARGS otherwise.
+StatusCode queue_init(Queue *queue);
```
mitchellostler, line 18: @ryandancy @JarvisWeng maybe it does make sense to have a wrapper here to declare all the static memory? Since we know what the size of the buffer will be from the struct params. 
We could do something like
```
#define queue_init(queue) \
static uint8_t s_##queue_buf[(&queue->size)*(&queue->item_size)]; \
&queue->storage_buf = s_##queue_buf; \
 _queue_init(&queue); 

```
I know we were trying to get away from macro magic, but this could make things cleaner in the projects?

mitchellostler: Ignore this comment, we can just have people declare their own buffers

mitchellostler:
Just the two nit items from me, otherwise you are good to go

mitchellostler (approved):
Just went ahead and added a couple comments so I can go ahead and get this merged. Great work!

hewittmcg:
@mitchellostler realized that the change from `Queue.size` to `Queue.num_items` caused a couple compilation errors, but should be good to go now :D

ryandancy (approved):
Might want to run it through `scons format` once (spaces in the if statement) but looks good 👍

mitchellostler:
> @mitchellostler realized that the change from `Queue.size` to `Queue.num_items` caused a couple compilation errors, but should be good to go now :D

Yup, sorry for breaking your code!

mitchellostler:
Sweet! good work getting this done

PR #25, opened 2022-03-05
Soft 595 can pack transmit struct

123Logan321:


libraries/codegen/templates/_pack.h.jinja
```diff
@@ -0,0 +1,27 @@
...
+{%- set signal_length = data["Messages"][message]['signals'][signal[0]].values() | list %}
+
+{% set message_1 -%}
+void can_pack_{{message}}(CanMessage *msg_ptr{%- for key in signal-%}, uint{{signal_length[0]}}_t {{key}}_u{{signal_length[0]}}{%- endfor%}){ {%- endset -%}
+
+{%- set message_2 %}
+can_pack_impl_u{{signal_length[0]}}((msg_ptr), SYSTEM_CAN_DEVICE_BMS_CARRIER, SYSTEM_CAN_MESSAGE_{{message|upper}}, {{data['Messages'][message]['id']}}, {%- endset -%}
+
+{%- set message_3 %}
+{% for key in signal%}{%if loop.first %} ({{key}}_u{{signal_length[0]}}){% else %}, ({{key}}_u{{signal_length[0]}}){% endif %}{% endfor-%}{% endset -%}
```
JarvisWeng, line 19: Indent this for please. It's inside another for

libraries/codegen/templates/_pack.h.jinja
```diff
@@ -0,0 +1,27 @@
...
+
+{%- set message_4 -%}
+{% for i in range((64 / signal_length[0] - data['Messages'][message]['signals']|length)| int )%}, CAN_PACK_IMPL_EMPTY{% endfor %}); }{%- endset -%}
+
+{% set bms_function %}{{message_1}}{{message_2}}{{message_3}}{{message_4}}{% endset -%}
```
JarvisWeng, line 24: Change this name

libraries/codegen/templates/_pack.h.jinja
```diff
@@ -0,0 +1,27 @@
...
+#include "can_msg_defs.h" 
+#include "can_pack_impl.h" 
+
+{% set messages = data["Messages"] -%}
+
+{%- for message in messages -%}
+{%- set signal = data['Messages'][message]['signals'].keys() | list -%}
+{%- set signal_length = data["Messages"][message]['signals'][signal[0]].values() | list %}
+
+{% set message_1 -%}
```
JarvisWeng, line 12: You should rename all the `message_1, message_2, ...` Those aren't really what they represent and make reading this more confusing

libraries/codegen/templates/_pack.h.jinja
```diff
@@ -0,0 +1,27 @@
...
+void can_pack_{{message}}(CanMessage *msg_ptr{%- for key in signal-%}, uint{{signal_length[0]}}_t {{key}}_u{{signal_length[0]}}{%- endfor%}){ {%- endset -%}
```
JarvisWeng, line 13: Try to put the `{%- endset -%}` on a new line. Makes the jinja more readable

libraries/codegen/generator.py
```diff
@@ -9,6 +9,57 @@ def read_yaml(yaml_file):
         data = yaml.load(f, Loader=yaml.FullLoader)
         return data
 
+def format_pack_h(env, template_name, file_path, data):
```
JarvisWeng, line 12: So instead of making a jinja filter, you made this function to format?

JarvisWeng (changes requested):

libraries/codegen/generator.py
123Logan321: Yeah I thought that is how you would format it, I will fix this error sorry.

JarvisWeng: Yeah, I would like it to be a Jinja filter, that way it's more portable between templates if we create more that need it

123Logan321: ok I will get to fixing the errors, thanks

libraries/codegen/generator.py
```diff
@@ -17,6 +17,37 @@ def read_yaml(yaml_file):
...
+    # Custom jinja filter that takes in the function parameter string
+    # and format it to indent to a new line whenever string length 
+    # reaches 100 characters.
+
+    template = template.split()
+    letter_count = 0
+    for i in range(len(template)):
+        letter_count += len(template[i]) + 1
+        if letter_count >= 100:
+            template[i] = "\n" + (" "*(len(template[0]) + len(template[1]) - 9)) + template[i]
```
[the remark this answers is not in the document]
123Logan321: Yeah, it is for adding the tabbing space, and template[0] always corresponds to "void" and template[1] is can_pack_name(CanMessage.

JarvisWeng (approved):

PR #26, opened 2022-03-05
Soft 50 can generator checker

Bafran:


libraries/codegen/generator.py
```diff
@@ -2,11 +2,15 @@
...
 import os
 import yaml
+import re
 
 
 def read_yaml(yaml_file):
     with open(yaml_file, "r") as f:
         data = yaml.load(f, Loader=yaml.FullLoader)
+        # Only check if a board is specified
+        if "Messages" in data.keys():
```
JarvisWeng, line 12: Should fail and raise exception or error out if doesn't exist

libraries/codegen/generator.py
```diff
@@ -33,8 +37,9 @@ def process_setter_data(board, data, master_data):
     for message in data["Messages"]:
         if board in data["Messages"][message].get("target", []) and \
            data["Messages"][message].get("signals", False):
-            for signal, length in data["Messages"][message]["signals"].items():
-                master_data["Signals"].append((signal, length['length']))
+            for signal, signal_data in data["Messages"][message]["signals"].items():
+                signal_data["message"] = message
```
JarvisWeng, line 41: Usually, I would advice against adding information to the `data`. This is because `data` should strictly come from the `yaml`, and since its added here, the orderly of function calls becomes important.
Though, I suppose its ok.

libraries/codegen/templates/_receive_structs.h.jinja
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
JarvisWeng, line 1: Sorry did you add this file?

libraries/codegen/templates/_setters.h.jinja
```diff
@@ -1,15 +1,13 @@
-#pragma once
```
JarvisWeng, line 1: If this is a merge issue, please rebase. Else if its another commit issue, then only include your own changes

libraries/codegen/generator.py
```diff
@@ -103,6 +108,34 @@ def parse_board_yaml_files(board):
 
     return master_data
 
+def check_yaml_file(data):
+    regex = re.compile('[@!#$%^&*()<>?/\|}{~:]')
+    message_ids = set()
+
+    for message in data["Messages"]:
+        # No same ids for messages within a message
+        if data["Messages"][message]["id"] in message_ids:
```
JarvisWeng, line 117: What if a message doesn't have an `id` keyword?

libraries/codegen/generator.py
```diff
@@ -103,6 +108,34 @@ def parse_board_yaml_files(board):
...
+            raise Exception("Duplicate message id")
+        # All ids are between 0-64
+        elif data["Messages"][message]["id"] > 64 or data["Messages"][message]["id"] < 0:
+            raise Exception("Invalid message id")
+        else:
+            message_ids.add(data["Messages"][message]["id"])
+        # No illegal characters in message names
+        if(regex.search(message) != None):
+                raise Exception("Illegal character in message name")
```
JarvisWeng, line 126: Extra indent?

JarvisWeng (changes requested):

PR #27, opened 2022-03-05
Update can_boards.h.jinja template to create message ids defines

AdrianJendo:
Can test with
`python3 libraries/codegen/generator.py -y libraries/codegen/boards/boards.yaml -t libraries/codegen/templates/can_board_ids.h.jinja`
and
`python3 libraries/codegen/generator.py -t libraries/codegen/templates/system_can.dbc.jinja`
to confirm output is correct


libraries/codegen/templates/can_board_ids.h.jinja
```diff
@@ -1,5 +1,111 @@
 #pragma once
 
+#define SYSTEM_CAN_MESSAGE_BPS_HEARTBEAT \
```
JarvisWeng, line 3: Sorry, these names need to be autogenerated not hard coded.
They will be the names of the
```
SYSTEM_CAN_MESSAGE_<message name>
(<signal name> << 5) + SYSTEM_CAN_DEVICE_<board>
```
Something like this

JarvisWeng (changes requested):

libraries/codegen/templates/can_board_ids.h.jinja
AdrianJendo: Might still be wrong, but let me know if it's on the right track

libraries/codegen/generator.py
```diff
@@ -36,10 +36,10 @@ def process_setter_data(board, data, master_data):
 
 def get_dbc_data():
     yaml_files = get_yaml_files()
-    master_data = {"Messages" : []}
-    for file in yaml_files:
-        data = read_yaml(file)
-        sender = get_board_name(file)
+    master_data = {"Messages" : [], "Boards": get_boards()}
+    for can_board_id, board in enumerate(master_data["Boards"]):
```
JarvisWeng, line 40: Might want to make `can_board_id` + 1. Not sure if we want to start at 0. But that's ok for now

libraries/codegen/templates/can_board_ids.h.jinja
```diff
@@ -1,13 +1,23 @@
+{% set boards = data["Boards"] -%}
+
 #pragma once
 
-#include <stdbool.h>
+{%- for board in boards -%}
```
JarvisWeng, line 5: All this needs to be below line 23. Since we want `SYSTEM_CAN_DEVICE....` and that's only defined below. Also just move the `#include` back at the top

libraries/codegen/generator.py
AdrianJendo: Would you prefer I change it still?

JarvisWeng: No that's ok. I'll change it later if its an issue

JarvisWeng (approved):

PR #28, opened 2022-03-05
[FWXV-13] Task testing helper library

ryandancy:
It's a helper library for creating task tests, which are tests that run in the context of a highest-priority task.

We define an `MS_TEST` preprocessor symbol when in test mode now, and we use this in FreeRTOSConfig.h to add a new highest priority so that the test task is always the highest priority. Also, users can redefine the max priority using the `NUM_FREERTOS_PRIORITIES` preprocessor symbol.


libraries/ms-freertos/inc/FreeRTOSConfig.h
```diff
@@ -1,17 +1,35 @@
...
 #include <stdint.h>
 extern uint32_t SystemCoreClock;
 #endif
 
+// Allow projects to add more priorities with the NUM_FREERTOS_PRIORITIES macro.
+// We can't have fewer than the default in case libraries depend on having enough priorities.
+#define DEFAULT_NUM_FREERTOS_PRIORITIES 5
+#ifdef NUM_FREERTOS_PRIORITIES
+#if NUM_FREERTOS_PRIORITIES < DEFAULT_NUM_FREERTOS_PRIORITIES
+#error Too few FreeRTOS priorities! We need at least DEFAULT_NUM_FREERTOS_PRIORITIES.
```
mitchellostler, line 14: Could we add something like this to the Task library to validate inputs (probably on a different branch)? Would make debugging faulty initialization a lot quicker, especially to help out newer members

ryandancy: I don't think we can autogenerate the `#error`s, it would be nice but I'm not sure it's possible to generate compiler errors for invalid initialization :(

libraries/ms-freertos/src/static_alloc.c
```diff
@@ -0,0 +1,37 @@
...
+  // state will be stored.
+  *ppxIdleTaskTCBBuffer = &xIdleTaskTCB;
+
+  // Pass out the array that will be used as the Idle task's stack.
+  *ppxIdleTaskStackBuffer = uxIdleTaskStack;
+
+  *pulIdleTaskStackSize = configMINIMAL_STACK_SIZE;
+}
+
+void vApplicationGetTimerTaskMemory(StaticTask_t **ppxTimerTaskTCBBuffer,
```
mitchellostler, line 25: This is an aside, but was there a major reason we were thinking of not using the provided soft_timer API? As part of ShiCheng's interrupts ticket we discussed making this part of the Gpio_it library, but it seems like FreeRTOS supports pretty much all we want to cover, where you can register callbacks and execute them in the context of this task.

mitchellostler (approved):
Looks good to me, just a couple small questions

JarvisWeng:
It makes sense to me. I think we just need to use more just to see how it feels and any additional features we may want. I think some documentation would be nice, like when someone would want  to create a TASK_TEST, as opposed to just writing stuff within the test. The macros might get confusing especially for new members

JarvisWeng (approved):

ShiCheng-Lu (commented):
looks good otherwise.

libraries/ms-freertos/inc/FreeRTOSConfig.h
```diff
@@ -1,17 +1,35 @@
...
+#define NUM_FREERTOS_PRIORITIES DEFAULT_NUM_FREERTOS_PRIORITIES
+#endif
+
+// Add an additional highest priority in test mode for the test task (see task_test_helpers.h)
+#ifdef MS_TEST
+#define ADDITIONAL_FREERTOS_PRIORITIES 1
+#else
+#define ADDITIONAL_FREERTOS_PRIORITIES 0
+#endif
+#define configMAX_PRIORITIES (NUM_FREERTOS_PRIORITIES + ADDITIONAL_FREERTOS_PRIORITIES)
```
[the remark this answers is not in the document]
ryandancy: There's a limit of 32 priorities total - I'll add some extra guarding for that.

libraries/ms-freertos/src/static_alloc.c
ryandancy: Hm, yeah it seems like we should use FreeRTOS's soft timer implementation, it's optimized better than we could do.  We'll probably end up just writing a wrapper over it then. I don't think there's a reason not to use it.

ryandancy:
@mitchellostler On why we need to hack around with the C preprocessor:

Unity's test runner generator, `generate_test_runner.rb`, is the script that generates a test runner with a `main` function from the test files we give it. The script is rather dumb, and it just goes through the test file looking for things that look like functions that start with `test_`. (In previous versions it would even get tripped up on whitespace! It's slightly better now but still dumb - for example, it treats `}` as the end of a line, which means you can't use arrays in parameterized tests, but that's for another time.)

Since it does everything syntactically, it doesn't expand C macros on its own, so the test functions I was generating with the helper library wouldn't work since `TEST_TASK(foo, 512)` doesn't look like a function that starts with `test_` to the runner generator script. So I had to pipe it through the C preprocessor (which can be done with `gcc -E`).

After that, I got *more* errors. It turns out that `gcc -E` will also expand all of the `#include` directives, including `#include "unity.h"`, which contains prototypes for `setUp()` and `tearDown()`. The runner generator script thought those were definitions of setup/teardown functions, so it was calling them instead of `setup_test` and `teardown_test` and causing linker errors. So to fix it I wrote the `preprocessor_filter.py` script which just filters out the content of any `#include`'d files from the preprocessor output using the [linemarkers](https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html) that the preprocessor generates. This results in a decently clean preprocessed file with the macros expanded that we can then feed into the runner generator script.

libraries/ms-common/inc/task_test_helpers.h
```diff
@@ -0,0 +1,26 @@
...
+
+#define TASK_TEST(test_name, stack_size)                                                  \
+  DECLARE_TASK(test_name);                                                                \
+  _TEST_TASK_IMPL(test_name);                                                             \
+  static void _prv_task_test_##test_name(void);                                           \
+  TASK(test_name, stack_size) {                                                           \
+    _prv_task_test_##test_name();                                                         \
+    vTaskEndScheduler(); /* Automagically stop FreeRTOS when the task function returns */ \
+  }                                                                                       \
+  static void _prv_task_test_##test_name(void)
```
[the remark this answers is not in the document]
ryandancy: The one issue with using a single task for all tests is that then the stack size is the same for all tests, which might be an issue if one test needs a larger stack size. Though you're right, we may run into issues with running out of memory if we run a test file on stm32 with too many task tests. Since it's a bit tricky to implement, I think it's okay to leave it for now and come back if it becomes an issue.

ryandancy:
@Kyle02 Sorry for the wait, you should be unblocked now :)

PR #29, opened 2022-03-08
FWXV-66: New Smoke Directory

danielq987:
JIRA: https://uwmidsun.atlassian.net/browse/FWXV-66

## Feature

Added the smoke directory in the root directory to store smoke tests.

## Buildsystem usage changes:
```bash
# New Project
scons new_smoke --project=<smoke_test_name>

# Building (same usage as regular project)
scons <smoke_test_name>

# Testing (same usage as regular project)
scons test --project=<smoke_test_name>

# sim, flash, gdb
scons sim_smoke --project=<smoke_test_name> --platform=x86
scons gdb_smoke --project=<smoke_test_name> --platform=x86
scons flash_smoke --project=<smoke_test_name> --platform=arm

# Linting/formatting (same usage as regular project)
scons --project=<smoke_test_name>
```


SConstruct
```diff
@@ -318,16 +321,23 @@ Alias('test', test)
 def make_new_target(target, source, env):
     # No project or library option provided
     if not PROJECT and not LIBRARY:
-        print("Missing project or library name. Expected --project=... or --library=...")
+        print("Missing project or library name. Expected --project=..., or --library=...")
         sys.exit(1)
+    
+    if str(target[0]) == 'new_smoke.txt' and PROJECT:
```
ryandancy, line 327: I'm not too keen on using the dummy filenames we pass to `Command()` for logic purposes - they're kind of a hack themselves and I'd like to get rid of them entirely at some point. I think there's an alternative though: according to [the `Command` docs](https://scons.org/doc/2.1.0/HTML/scons-user/a10706.html#f-Command) you can pass keyword arguments to `Command` that are mapped to environment variables.

I think then that if you pass `smoke=True` as an argument to the smoke `Command()` calls (and `smoke=False` to their non-smoke counterparts), then you should be able to use `env['smoke']` to determine whether you're in the smoke version of the command or not. That should save testing against the ugly `.txt` files.

ryandancy (changes requested):
Looks great, seems like it works! Just one cleanliness thing and then we should be good to go.

PR #30, opened 2022-03-12
Soft 605 log library

Kyle02:
Not yet ready to merge, still need to add tests. Just setting up a pr for comments on the implementation.


libraries/core/src/log.c
```diff
@@ -0,0 +1,31 @@
...
+#define ITEM_SIZE sizeof(g_buffer)
+
+// Holds the output of snprintf
+char g_buffer [MAX_LOG_SIZE];
+
+// s_xQueueBuffer will hold the queue structure.
+static StaticQueue_t s_xQueueBuffer;
+
+// Array to hold items
+static uint8_t s_log_array[QUEUE_LENGTH * ITEM_SIZE] = {0};
```
mitchellostler, line 12: the `= {0}` here is redundant, static declarations already zero the memory.

mitchellostler (changes requested):
Looks really good! I might have gone a bit overboard with the comments, but almost everything is superficial, all the core structure is great. 

You'll need to rebase on master, and I'll get the queue library merged so you can use that in this program as well.

mitchellostler:
There's also this merge conflict that you'll have to resolve, but I don't think it will be a tough fix.

libraries/core/inc/log.h
```diff
@@ -1,31 +1,57 @@
...
+    else                                                                                                                \
+      xQueueSendToBack( g_log_queue, ( void * ) g_buffer, ( TickType_t ) 0 );                                           \
+  } while (0)
+
+#define LOG_C(level, fmt, ...)                                                                                          \
+  do {                                                                                                                  \
+    snprintf(g_buffer, MAX_LOG_SIZE, "[%u] %s:%u: " fmt, (level), __FILE__, __LINE__, ##__VA_ARGS__);                   \
+    if(xTaskGetSchedulerState() == taskSCHEDULER_SUSPENDED || xTaskGetSchedulerState() == taskSCHEDULER_NOT_STARTED)    \
+      printf("%s", g_buffer);                                                                                           \
+    else{                                                                                                               \
```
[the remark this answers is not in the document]
Kyle02: it seems that scons format does catch format errors in macros, which is good to know

libraries/ms-common/inc/log.h
```diff
@@ -0,0 +1,73 @@
...
+#define MAX_LOG_SIZE (size_t)200
+
+#ifndef IN_ORDER_LOGS
+#define IN_ORDER_LOGS 0
+#endif
+
+#define MIN_TASK_PRIORITY 1
+
+extern QueueHandle_t g_log_queue;
+extern char g_buffer[MAX_LOG_SIZE];
```
ShiCheng-Lu, line 24: I'd suggest renaming this to g_log_buffer just in case we have other global buffers.

libraries/ms-common/src/log.c
```diff
@@ -0,0 +1,37 @@
...
+// use it)
+QueueHandle_t g_log_queue = NULL;
+
+void log_init(void) {
+  g_log_queue = xQueueCreateStatic(QUEUE_LENGTH, ITEM_SIZE, s_log_array, &s_log_queue);
+  tasks_init_task(log_task, TASK_PRIORITY(MIN_TASK_PRIORITY), NULL);
+}
+
+TASK(log_task, TASK_STACK_256) {
+  char rx_buffer[MAX_LOG_SIZE];
```
ShiCheng-Lu, line 24: I think it's better for this rx_buffer to be static so the compiler knows how much space our program is using. I'm not sure how it works exactly if it can be static in function, static at top level, or left like this. (ask Mitchell or Ryan I'm sure one of them knows).

libraries/ms-common/src/log.c
```diff
@@ -0,0 +1,37 @@
...
+
+  // All tasks MUST loop forever and cannot return.
+  while (true) {
+    if (uxQueueSpacesAvailable(g_log_queue) == 0) {
+      printf("Error: Log queue is full\n");
+    }
+    if (uxQueueMessagesWaiting(g_log_queue) != 0) {
+      if (xQueueReceive(g_log_queue, (rx_buffer), (TickType_t)0)) {
```
ShiCheng-Lu, line 32: why not just xQueueReceive with a long time instead of checking MessagesWaiting != 0?

libraries/ms-common/src/log.c
mitchellostler: It can be static at top level, or static in the function. Either is fine, and mean the memory will only be initialized once. I think it could be a good idea to leave it in the function here, since it isn't used anywhere else.

libraries/ms-common/inc/log.h
```diff
@@ -0,0 +1,71 @@
...
+#define QUEUE_LENGTH 32
+#define MAX_LOG_SIZE (size_t)200
+
+#ifndef IN_ORDER_LOGS
+#define IN_ORDER_LOGS 0
+#endif
+
+#define MIN_TASK_PRIORITY 1
+
+extern QueueHandle_t g_log_queue;
```
mitchellostler, line 23: Add comment for global handle `global queue handle used in all log calls` or something like that

libraries/ms-common/src/log.c
```diff
@@ -0,0 +1,36 @@
...
+// Holds the output of snprintf
+char g_log_buffer[MAX_LOG_SIZE];
+
+// s_log_queue will hold the queue structure.
+static StaticQueue_t s_log_queue;
+
+// Array to hold items
+static uint8_t s_log_array[QUEUE_LENGTH * ITEM_SIZE];
+
+// Holds the queue object (either make this into the task or put the variables in header so task can
```
mitchellostler, line 14: This comment is a bit weird. Maybe just say `queue handle definition`

mitchellostler (approved):

PR #32, opened 2022-03-13
Fwxv_73_soft_timer

ShiCheng-Lu:
Implement soft timer and watchdog as freeRTOS timer wrappers


libraries/ms-common/inc/soft_timer.h
```diff
@@ -0,0 +1,40 @@
...
+  StaticTimer_t buffer;
+  TimerHandle_t handle;
+} SoftTimer;
+
+typedef void (*SoftTimerCallback)(TimerHandle_t handle);
+
+// Adds a software timer. The provided duration is the number of
+// miliseconds before running and the callback is the process to run once
+// the time has expired.
+StatusCode soft_timer_start(uint32_t duration_ms, SoftTimerCallback callback, SoftTimer *timer);
```
JarvisWeng, line 26: Do we not allow context? Or just assume global variables to be used in the callback

ShiCheng-Lu: TBH, I excluded it because I have never (never) seem context being used, and it's almost always the case that the callback is in the same file as the static items it needs to access. It is also a lot more complicated to add a context field to the callback since the callback is called by FreeRTOS, so we'll need to let FreeRTOS call some callback that calls our own callback with a context. We can also get around this by using the pvTimerID field which is essentially just a free pointer that can be accessed via vTimerSetTimerID() and pvTimerGetTimerID().

libraries/ms-common/src/soft_timer.c
```diff
@@ -0,0 +1,51 @@
+#include "soft_timer.h"
+
+StatusCode soft_timer_start(uint32_t duration_ms, SoftTimerCallback callback, SoftTimer *timer) {
+  return soft_timer_start_with_context(duration_ms, callback, timer, NULL);
+}
+
+StatusCode soft_timer_start_with_context(uint32_t duration_ms, SoftTimerCallback callback,
```
mitchellostler, line 7: @ryandancy what are your thoughts on this implementation for using context in the soft-timer functions? 
If we can get away with not using context, then that would be cleaner. Not sure how often we needed to use context in IV

mitchellostler: Ok, I talked with Ryan. 
We don't need context for our soft_timers for now. Realistically we shouldn't be passing any shared data to the soft_timer task, as it creates a potential race condition (which is probably why they don't provide a context param).
Generally, if periodic behaviour needs to happen with shared data, it should just send a notification to the task, and the callback should operate in the context of the task which started the timer. 
We may add a soft_timer_start_notify() wrapper at a later date, but for now this is good.

libraries/ms-common/src/soft_timer.c
```diff
@@ -0,0 +1,40 @@
...
+  timer->id = xTimerCreateStatic(NULL, pdMS_TO_TICKS(duration_ms), pdFALSE,  //
+                                 NULL, callback, &timer->buffer);
+  if (xTimerStart(timer->id, 0) != pdPASS) {
+    return status_msg(STATUS_CODE_INTERNAL_ERROR, "timer command queue is full");
+  }
+  return STATUS_CODE_OK;
+}
+
+void *soft_timer_get_context(SoftTimerId id) {
+  return pvTimerGetTimerID(id);
```
JarvisWeng, line 13: Is this a private function or what does this do?

JarvisWeng (approved):

libraries/ms-common/src/soft_timer.c
ShiCheng-Lu: I forgor

PR #38, opened 2022-03-31
Fwxv 74 queues library tests

Bafran:


libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+#include "queues.h"
+#include "unity.h"
+
+#define LIST_SIZE 5
+#define ITEM_SIZE sizeof(uint8_t)
+
+#define QUEUE_LENGTH 5
+#define BUF_SIZE (QUEUE_LENGTH * ITEM_SIZE)
+
+static const uint8_t s_list[LIST_SIZE] = {
```
JarvisWeng, line 16: Nit: Put all on 1 line

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+	5, 6, 7, 8, 9
+};
+
+static StaticQueue_t s_queue;
```
JarvisWeng, line 20: Why do you declare these variables?

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+static QueueHandle_t s_queue_handle;
+
+static Queue queue;
+
+static bool s_task_started;
+
+void setup_test(void) {
+  s_task_started = false;
+
+  queue.handle = s_queue_handle;
```
JarvisWeng, line 31: You're misunderstanding the usage of the `queues.h` You shouldn't set the handler or queue. Only the item size, num of items and storage pointer.

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+  uint8_t index = 0;
+  volatile TickType_t tick_count;
+
+  while (true) {
+    do {
+      status = queue_send(&queue, (void *) &s_list[index], 100);
+      tick_count = xTaskGetTickCount();
+      if (status==STATUS_CODE_OK)
+        LOG_DEBUG("Sent int: %d at tick %lu\n", s_list[index], tick_count);
+      if (++index == 5) {
```
JarvisWeng, line 57: Nit: Replace 5 with `QUEUE_LENGTH`

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+    delay_ms(200);
+    while (true) {
+      status = queue_receive(&queue, &outstr, 100);
+      tick_count = xTaskGetTickCount();
+      if (status != STATUS_CODE_OK) {
+        LOG_DEBUG("Error reading from queue at tick %lu\n", tick_count);
+        delay_ms(100);
+        continue;
+      }
+      LOG_DEBUG("Received int: %d at tick %lu\n", outstr, tick_count);
```
JarvisWeng, line 80: Logs are ok for debugging, but you should use `TEST_ASSERT_TRUE` or something to test this out

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+TASK(peekMessages, TASK_STACK_512) {
+    BaseType_t status;
+    const uint8_t outstr = 0;
+    volatile TickType_t tick_count;
+
+    while(true) {
+      status = queue_peek(&queue, &outstr, 200);
+      tick_count = xTaskGetTickCount();
+      if (status != STATUS_CODE_OK) {
+        LOG_DEBUG("Failed to peek at tick %lu\n", tick_count);
```
JarvisWeng, line 93: Again use `TEST_ASSERT`s

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,132 @@
...
+      } else {
+        LOG_DEBUG("Peeked item %d at %lu\n", outstr, tick_count);
+        delay_ms(500);
+      }
+    }
+}
+
+TASK_TEST(test_running_task, TASK_STACK_512) {
```
JarvisWeng, line 101: In this test, you should be asserting the QUEUE contents, not the status. Probably be easier in the `receiveMessages` task or peek instead.

JarvisWeng (changes requested):

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,96 @@
...
+void teardown_test(void) {}
+
+TASK(sendMessages, TASK_STACK_512) {
+  BaseType_t status;
+  uint8_t index = 0;
+
+  while (true) {
+    do {
+      status = queue_send(&queue, (void *) &s_list[index], 100);
+      if (status==STATUS_CODE_OK)
```
JarvisWeng, line 41: Enclose this properly. What do this single line `if` suppose to do?

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,96 @@
...
+    BaseType_t status;
+    uint8_t outstr = 0;
+    uint8_t recv_str[QUEUE_LENGTH] = {0};
+    uint8_t index = 0;
+
+    delay_ms(200);
+    while (true) {
+      status = queue_receive(&queue, &outstr, 100);
+      if (status != STATUS_CODE_OK) {
+        TEST_ASSERT_EQUAL_UINT8_ARRAY(s_list, recv_str, QUEUE_LENGTH);
```
JarvisWeng, line 60: You should assert the contents regardless if status is not `STATUS_CODE_OK`. In fact, under normal circumstances, it will never go into this `if` statement. Since it `status != STATUS_CODE_OK` will only happen if it doesn't receive. So the recv str, should be different anyway

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,96 @@
...
+
+TASK(peekMessages, TASK_STACK_512) {
+    BaseType_t status;
+    uint8_t outstr = 0;
+
+    while(true) {
+      status = queue_peek(&queue, &outstr, 200);
+      if (status != STATUS_CODE_OK) {
+      } else {
+        TEST_ASSERT_EQUAL_UINT8(5, outstr);
```
JarvisWeng, line 79: Why's it always 5?

libraries/ms-common/test/test_queues.c
```diff
@@ -0,0 +1,99 @@
...
+      status = queue_receive(&queue, &outstr, 100);
+      if (status != STATUS_CODE_OK) {
+        // Outstrs are the same for previous and current iteration,
+        // recieve failed
+        TEST_ASSERT_EQUAL_UINT8(outstr, prev_outstr);
+        delay_ms(100);
+        continue;
+      }
+      // Outstr is expected value
+      TEST_ASSERT_EQUAL_UINT8(outstr, s_list[index]);
```
JarvisWeng, line 65: Shouldn't this assert be in an `else`?

Bafran: I've used a continue in the previous if statement which makes sure it's reading the correct status, but I'll change it to an else to make it more readable.

JarvisWeng (approved):

PR #39, opened 2022-04-02
Fwxv 41 can transmit struct template

123Logan321:


libraries/codegen/templates/_tx_structs.h.jinja
```diff
@@ -0,0 +1,16 @@
...
+{%- set board = data["Board"] %}
+
+#include <stdint.h>
+
+typedef struct
+{
+{%- for message in messages -%}
+    {%- set signals = data['Messages'][message]['signals'].keys() | list -%}
+    {%- for signal in signals %}
+    uint8_t {{message}}_{{signal}};
```
JarvisWeng, line 13: The size isn't always necessarily a `uint8_t` It can also be a `uint16_t`, `uint32_t`, or `uint64_t`. But that's ok. I'll just merge it and fix it up myself

JarvisWeng (approved):

PR #41, opened 2022-04-03
Fwxv 70 thread safe gpio

AdrianJendo:
Test with `scons test --library=ms-common --platform=x86 --test=gpio`


libraries/ms-common/src/arm/gpio.c
```diff
@@ -7,20 +7,33 @@
 #include "stm32f0xx.h"
 #include "stm32f0xx_gpio.h"
 #include "stm32f0xx_rcc.h"
+#include "mutex.h"
 
 static GPIO_TypeDef *s_gpio_port_map[] = { GPIOA, GPIOB, GPIOC, GPIOD, GPIOE, GPIOF };
 static uint32_t s_gpio_rcc_ahb_timer_map[] = { RCC_AHBPeriph_GPIOA, RCC_AHBPeriph_GPIOB,
                                                RCC_AHBPeriph_GPIOC, RCC_AHBPeriph_GPIOD,
                                                RCC_AHBPeriph_GPIOE, RCC_AHBPeriph_GPIOF };
+static Mutex s_gpioMutex;
```
JarvisWeng, line 16: Nit: Usually we stick with 1 naming convention, so change it to `s_gpio_mutex` please

libraries/ms-common/src/arm/gpio.c
```diff
@@ -7,20 +7,33 @@
...
 
 StatusCode gpio_init(void) {
-  return STATUS_CODE_OK;
+  StatusCode status = mutex_init(&s_gpioMutex);
```
JarvisWeng, line 19: If there's only 1 mutex, then a user can only access 1 GPIO at a time. Not really an issue since we only have 1 GPIO but ideally there should be a mutex for each GPIO rather than just 1.

libraries/ms-common/src/arm/gpio.c
```diff
@@ -66,21 +79,33 @@ StatusCode gpio_init_pin(const GpioAddress *address, const GpioSettings *setting
 
   // Use the init_struct to set the pin.
   GPIO_Init(s_gpio_port_map[address->port], &init_struct);
+  mutex_unlock(&s_gpioMutex);
   return STATUS_CODE_OK;
 }
 
 StatusCode gpio_set_state(const GpioAddress *address, GpioState state) {
-  if (address->port >= NUM_GPIO_PORTS || address->pin >= GPIO_PINS_PER_PORT ||
+  StatusCode status = mutex_lock(&s_gpioMutex, BLOCK_INDEFINITELY);
```
JarvisWeng, line 87: This might be an issue say if a Task is accessing a GPIO and an interrupt comes in and it also needs to access the same GPIO.

JarvisWeng: To solve this, you might want to look into CRTICAL SECTIONS instead.
Or make it so it doesn't BLOCK_INDEFINITELY and make a timeout in case can't access GPIO

libraries/ms-common/test/test_gpio.c
```diff
@@ -0,0 +1,83 @@
...
+
+  // Loop forever, because tasks should not exit.
+  while (true) {
+    status = gpio_set_state(&s_gpio_addr, GPIO_STATE_LOW);
+    TEST_ASSERT_TRUE(status == STATUS_CODE_OK);
+    status = gpio_get_state(&s_gpio_addr, &state);
+    TEST_ASSERT_TRUE(status == STATUS_CODE_OK);
+    TEST_ASSERT_TRUE(state == GPIO_STATE_LOW);
+    LOG_DEBUG("task1 loop finished\n");
+    delay_ms(300);
```
JarvisWeng, line 40: You shouldn't use `delay_ms`, this will cause this task to hold the resources indefinitely if its higher priority

libraries/ms-common/test/test_gpio.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+    LOG_DEBUG("task2 loop finished\n");
+    delay_ms(100);
+  }
+}
+
+TASK_TEST(test_running_task, TASK_STACK_512) {
+  // Start the task: note no need to call tasks_start() because we're inside a test task and
+  // FreeRTOS is already running.
+  tasks_init_task(task1, TASK_PRIORITY(2), NULL); // higher priority
```
JarvisWeng, line 78: This test does not work.
What happens is that `task1` will continuously run and never give up resources. `task2` should never have a chance to run. This will end after 5000ms

JarvisWeng: You need to write the test such that it tests the GPIO operations between tasks. So there's 2 cases: different priority tasks, and same priority tasks.

So for different priority tasks
Ex: Task1 P1, Task2 P2
- Task1:
     - set GPIO HIGH. Non-blocking delay 10ms
     - get GPIO assert HIGH. Non-blocking delay 10ms
     - toggle GPIO. Non-blocking delay 10ms
     - get GPIO. Assert LOW. Blocking delay 10ms <- Only now does task 1 give up resources
     - ...other stuff
- Task2:
     - get GPIO and assert LOW. Non-blocking delay 5ms <- Happens after task1 runs
     - toggle GPIO. Non-blocking delay 5ms <- Now task1 can interrupt it again since higher priority and 10ms have passed
     - get GPIO. Assert LOW again
     - ...other stuff

If it helps draw a diagram with each task side by side and see the order of operations

JarvisWeng (changes requested):

libraries/ms-common/test/test_gpio.c
```diff
@@ -0,0 +1,93 @@
...
+    TEST_ASSERT_TRUE(status == STATUS_CODE_OK);
+    TEST_ASSERT_TRUE(state == GPIO_STATE_LOW);
+    usleep(10000);
+
+    status = gpio_toggle_state(&s_gpio_addr);
+    TEST_ASSERT_TRUE(status == STATUS_CODE_OK);
+    usleep(10000);
+
+    status = gpio_get_state(&s_gpio_addr, &state);
+    TEST_ASSERT_TRUE(status == STATUS_CODE_OK);
```
JarvisWeng, line 76: Technically `usleep` and `delay_ms` use different clocks, so it's not really exact. But alright for now.
Surprised this ran.
On line 70, `task 1` should start again and run till completion,
and then `task 2` starts again and toggles the GPIO.
Then `task 1` should start again and GPIO should be `LOW`
So technically, `state == GPIO_STATE_LOW` again

JarvisWeng: To be more sure, change up the delays like 5000 us so it'll be sure of which tasks are together.

JarvisWeng (approved):

PR #42, opened 2022-04-26
FWXV-68: Use mutexes in retarget.c for logging

danielq987:
Use mutex in _write instead of disabling interrupts.

I'm not sure how to test this? If there are ways please comment :)

EDIT: Logging was made thread-safe, so the use of mutexes was not necessary. Simply disabling interrupts should not result in any issues, cc Mitchell.


mitchellostler (approved):
L G T M 💯

PR #44, opened 2022-05-18
Fwxv 61 integrate sanitizers

123Logan321:


SConstruct
```diff
@@ -55,6 +63,9 @@ LIBRARY = GetOption('library')
 ###########################################################
 
 # Retrieve the construction environment from the appropriate platform script
+if SANITIZER != 'none':
```
mitchellostler, line 66: Can you move all of the sanitizer logic, including when you parse the asan vs tsan to just below the "Add flags when compiling a test". Would be good to have it in one spot, and I think it makes more sense to be there. 

Other than that, looks good 👍

mitchellostler (changes requested):
If you could just move all the new functionality into the one place I indicated, otherwise I think you are good to go

mitchellostler (approved):

PR #45, opened 2022-05-18
Fwxv 72 port uart library

Bafran:


Bafran:
"stm32f0xx_interrupt.h" includes stm32f0xx_interrupt_nvic_enable. USART1_IRQn are undefined but included in header file "stm32f0xx.h"

mitchellostler:
Hey, good work getting this done. The implementation I had in my mind was a bit different than what you've done here,  and then I spent some more time reading through UART, and now I have a better idea to help the interface make more sense. 

Basically, what I'm thinking is that the entire UART interface will be done through queues, and uart tx/rx will essentially be wrappers over top of queue methods.

For each port, there will be 2 byte queues, rx and tx, both with uart_max_size. 

`uart_tx(port, u8 *data, size *len)`
- Will send (one byte at a time) data into the port's tx queue, with queue_send() in a for loop. The timeout should be zero, and if the operation fails (queue is full) that should be returned. Len should be set to number of bytes sent
- Will also initialize the uart_tx interrupt behaviour for transmission
- The tx interrupt handler will pop a byte of data using queue_receive_fromISR() freertos method in the tx_and send it 

`uart_rx(port, u8 *data, *len)`
- This method will read from the rx queue into the buffer provided by data. Should receive up to len items, and len should be set to number of items read. If all items are read, return status code ok, otherwise return not ok. (do all your null checking and what not). 
- The rx interrupt handler will automatically read into the queue with queue_send_fromISR() method whenever it is called (time out immediately), and if the buffer is full it should be reset. 

I think I still need to read more into the interrupt semantics of UART and the std periph layer, but this is the general structure I'm thinking. Let's talk a bit about it before you go and implement it.

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,204 @@
...
+
+  gpio_init_pin(&settings->tx, &gpio_settings);
+  gpio_init_pin(&settings->rx, &gpio_settings);
+
+  USART_InitTypeDef usart_init;
+  USART_StructInit(&usart_init);
+  usart_init.USART_BaudRate = BAUDRATE;
+  USART_Init(s_port[uart].base, &usart_init);
+
+  USART_ClearITPendingBit(s_port[uart].base, USART_FLAG_TXE);
```
mitchellostler, line 118: If you've researched and determined exactly what this functionality is doing and why, it would be great to add comments. This goes for all of the USART_*  and stm32f0xx std periph library calls. The person who wrote the fwxiv version should have done this initially, It would have made it much easier to do this update

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,204 @@
...
+      break;
+    }
+  }
+
+  if (USART_GetFlagStatus(s_port[uart].base, USART_FLAG_TXE) == SET)
+    USART_ITConfig(s_port[uart].base, USART_IT_TXE, ENABLE);
+  return status;
+}
+
+StatusCode uart_rx(UartPort uart, uint8_t *data, size_t *len) {
```
mitchellostler, line 148: Should always check for NULL pointers, It's really bad if we ever de-reference them.

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,204 @@
...
+                    .rx_queue = &rx_queue_port_4,
+                    .tx_buf = tx_buf_port_2,
+                    .rx_buf = rx_buf_port_2,
+                    .irq = USART3_4_IRQn,
+                    .base = USART4 },
+};
+
+static void prv_handle_irq(UartPort uart);
+
+StatusCode uart_init(UartPort uart, UartSettings *settings) {
```
mitchellostler, line 87: Check for NULL here as well. Not as important as it's a initialization method, not one that will be actively called as we're running our program, but still good practice

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,179 @@
...
+  if (USART_GetITStatus(s_port[uart].base, USART_IT_TXE) == SET) {
+    uint8_t tx_data = 0;
+    // If tx queue is empty, disable tx interrupts and return
+    if (xQueueReceiveFromISR(&s_port_queues[uart]->tx_queue.handle, &tx_data, pdFALSE) == pdFALSE) {
+      // Disable TX Interrupts to end transmission
+      USART_ITConfig(s_port[uart].base, USART_IT_TXE, DISABLE);
+    } else {
+      USART_SendData(s_port[uart].base, tx_data);
+    }
+    USART_ClearITPendingBit(s_port[uart].base, USART_IT_TXE);
```
Bafran, line 147: Might be unnecessary to clear pending bit since it should be cleared when `USART_SendData` is called

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,179 @@
...
+  USART_TypeDef *base;
+  bool initialized;
+} UartPortData;
+
+static UartPortQueue queue_port_1;
+static UartPortQueue queue_port_2;
+static UartPortQueue queue_port_3;
+static UartPortQueue queue_port_4;
+
+static UartPortQueue *s_port_queues[NUM_UART_PORTS] = {
```
mitchellostler, line 29: Nit: just do this as an array of objects, not pointers

`static UartPortQueue s_port_queues[NUM_UART_PORTS];`

We can just implicitly assume that each one can be accessed via UART_PORT_X. This won't change your structure at all, you will just have to use s_port_queues[uart].xx instead of s_port_queues[uart]->xx

libraries/ms-common/inc/uart.h
```diff
@@ -0,0 +1,32 @@
...
+  GpioAddress rx;  // rx pin
+  GpioAltFn alt_fn;
+  uint32_t baudrate;
+} UartSettings;
+
+// Assumes standard 8 N 1
+// Port 1 is reserved for retarget.c
+StatusCode uart_init(UartPort uart, UartSettings *settings);
+
+// RX data is read to *data with length returned as *len
```
mitchellostler, line 25: I'm thinking the following:
```
// Reads |*len| bytes of data into |*data| from UART rx queue at port |port|, up to UART_MAX_BUFFER_LEN. 
// If less than |*len| available, it is updated to number of bytes read.
// If rx data buffer is full, oldest data is removed from front of queue.
// This method should not be called from more than one task for the same port
```

libraries/ms-common/inc/uart.h
```diff
@@ -0,0 +1,32 @@
...
+// *data read into and overwritten from ISR queue buffer when function is called
+StatusCode uart_rx(UartPort uart, uint8_t *data, size_t *len);
+
+// Non-blocking TX
```
mitchellostler, line 29:
```
// Sends |*len| bytes of data from |*data| UART tx queue at port |port|,
// up to UART_MAX_BUFFER_LEN or if queue is full.
// If less than |*len| sent, it is updated to total bytes sent.
// This method should not be called from more than one task for the same port
```

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,179 @@
...
+}
+
+StatusCode uart_rx(UartPort uart, uint8_t *data, size_t *len) {
+  if (data == NULL || len == NULL) return STATUS_CODE_INVALID_ARGS;
+  StatusCode status;
+  for (uint8_t i = 0; i < *len; i++) {
+    status = queue_receive(&s_port_queues[uart]->rx_queue, &data[i], 0);
+    if (status != STATUS_CODE_OK) {
+      *len = i;
+      break;
```
mitchellostler, line 129: Should return Status code incomplete

libraries/ms-common/src/uart.c
```diff
@@ -0,0 +1,179 @@
...
+  return STATUS_CODE_OK;
+}
+
+StatusCode uart_tx(UartPort uart, uint8_t *data, size_t *len) {
+  if (data == NULL || len == NULL) return STATUS_CODE_INVALID_ARGS;
+  StatusCode status;
+  for (uint8_t i = 0; i < *len; i++) {
+    status = queue_send(&s_port_queues[uart]->tx_queue, &data[i], 0);
+    if (status != STATUS_CODE_OK) {
+      *len = i;
```
mitchellostler, line 111: return status code incomplete after setting len, instead of breaking loop

mitchellostler (commented):
Couple more comments, but it's very close. I will aim to hw test it next Saturday

ShiCheng-Lu:
you can solve the build issues by adding 
`pip install --upgrade protobuf==3.20`
after line 80 of /.github/workflows/main.yml

libraries/ms-common/src/uart.c
Bafran: I changed it to update status to incomplete, but still break so that even if the queue gets full the TX still gets enabled and sends whatever was passed to the queue, but returns status code incomplete.

mitchellostler (approved):

PR #46, opened 2022-05-21
First commit

A3sanjay:
I wrote up the filter autogeneration script template.

### DIFF
--- a/libraries/codegen/templates/filters.h.jinja
+++ b/libraries/codegen/templates/filters.h.jinja
@@ -1,0 +1,19 @@
+#pragma once
+
+#include "can_boards_ids.h"
+#include "can_codegen.h"
+
+{% set board = data["Board"] -%}
+{% set messages = data["Messages"] -%}
+
+#include <stdint.h>
+
+typedef struct
+{
+{%- for message, message_data in messages.items() -%}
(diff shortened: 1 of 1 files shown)


libraries/codegen/templates/filters.h.jinja
```diff
@@ -0,0 +1,19 @@
...
+#include "can_boards_ids.h"
+#include "can_codegen.h"
+
+{% set board = data["Board"] -%}
+{% set messages = data["Messages"] -%}
+
+#include <stdint.h>
+
+typedef struct
+{
```
JarvisWeng, line 12: This shouldn't be wrapped in a struct.
We want to call these functions so they should be in a function. something like `add_all_can_filters`

libraries/codegen/templates/filters.h.jinja
```diff
@@ -0,0 +1,19 @@
...
+{%- for message, message_data in messages.items() -%}
+    {%- set ids = message_data['id'] -%}
+    {%- for id, id_data in ids.items() %}
+    can_add_filter({{message}}_{{id}});
```
JarvisWeng, line 16: This line should also add the can filter ids. You can find them in `can_boards_ids.h`

PR #50, opened 2022-05-22
Fwxv 69 adc

ShiCheng-Lu:


libraries/ms-common/inc/adc.h
```diff
@@ -0,0 +1,47 @@
...
+  NUM_ADC_MODES,
+} AdcMode;
+
+extern const GpioAddress ADC_BAT;
+extern const GpioAddress ADC_REF;
+extern const GpioAddress ADC_TEMP;
+
+// const GpioAddress ADC_CHANNEL_BAT = { .port = NUM_GPIO_PORTS, .pin = 15 };
+
+typedef void (*AdcCallback)(uint8_t adc_channel, void *context);
```
JarvisWeng, line 25: Can we not allow ADCs to have callbacks? I think ADC reading should just be periodically read from a task

libraries/ms-common/inc/adc.h
```diff
@@ -0,0 +1,47 @@
...
+#include <stdint.h>
+
+#include "gpio.h"
+#include "notify.h"
+#include "status.h"
+#include "tasks.h"
+
+typedef enum {
+  ADC_MODE_SINGLE = 0,
+  ADC_MODE_CONTINUOUS,
```
JarvisWeng, line 15: How about sampling mode instead of continuous mode?

libraries/ms-common/inc/adc.h
```diff
@@ -0,0 +1,47 @@
...
+
+// Enable or disable a given pin.
+// A race condition may occur when setting a pin during a conversion.
+// However, it should not cause issues given the intended use cases
+// To set adc channels REF/TEMP/BAT, you must use adc_set_channel() below
+StatusCode adc_set_channel(GpioAddress address, bool new_state);
+
+// Register a callback function to be called when the specified pin
+// completes a conversion
+StatusCode adc_register_event(GpioAddress address, Task *task, Event event);
```
JarvisWeng, line 38: I personally do not want these small modules to notify tasks other tasks directly.
It should be like just returns the value for it

libraries/ms-common/src/arm/adc.c
```diff
@@ -0,0 +1,297 @@
...
+  uint8_t channel;
+  status_ok_or_return(adc_get_channel(address, &channel));
+  status_ok_or_return(prv_check_channel_enabled(channel));
+
+  if (!s_adc_status.continuous) {
+    xSemaphoreTake(s_adc_status.converting, 0);
+    ADC_StartOfConversion(ADC1);
+
+    // wait until conversion is finished
+    xSemaphoreTake(s_adc_status.converting, portMAX_DELAY);
```
JarvisWeng, line 263: Wait, you take, then take, the give?

libraries/ms-common/inc/adc.h
ShiCheng-Lu: won't fix

libraries/ms-common/src/arm/adc.c
ShiCheng-Lu: yes, only one conversion at a time, and block this thread while conversion is happening.

libraries/ms-common/inc/adc.h
ShiCheng-Lu: won't fix

mitchellostler (approved):
Looks good, good work getting this done. The interface is now much nicer than XIV lmao. 
The tests look great too

PR #51, opened 2022-05-22
103 tasks

awaterston01:
Completed 103 
 - task 1: to implement tasks that log a counter value (tasks.c file)
 - task 2: to implement tasks the send to and read from a queue (main.c file)


mitchellostler (commented):
Looks good! I'll close this now, but you're good to go for firmware 103

PR #54, opened 2022-05-28
Fw 999 addison - fw103 onboarding task

addison-ch:


projects/freertosproj1/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+static uint8_t s_queue1_buf[BUF_SIZE];
+static Queue s_queue1 = {
+  // Add parameters
+  .num_items = QUEUE_LEN,
+  .item_size = ITEM_SZ,
+  .storage_buf = s_queue1_buf,
+};
+
+TASK(task1, TASK_STACK_512) {
+  LOG_DEBUG("Task 1 initialized!\n");
```
mitchellostler, line 30: Nit: You don't need this, this will log automatically when the task starts.

projects/freertosproj1/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+  StatusCode ret;
+  int i = 0;
+  while (true) {
+    ret = queue_send(&s_queue1, &s_list[i], 0);
+    if (ret == STATUS_CODE_OK) {
+      i = (i == 5) ? 0 : i + 1;
```
mitchellostler, line 36: The ternary syntax here is a bit over-complicated. It's better to just do this in a for loop inside the while true loop. It does the same thing, but the for loop is much more readable and makes more sense for the functionality we are trying to achieve

mitchellostler (commented):
Overall looks good! if you could just make that one change to a for loop, and resubmit, you will be good to go.

PR #55, opened 2022-05-29
Fwxv 999 harsh 103 hw

harsh-patell:


projects/fw103/src/main.c
```diff
@@ -0,0 +1,72 @@
...
+  .item_size = ITEM_SZ,
+  .storage_buf = s_queue1_buf,
+};
+
+TASK(task1, TASK_STACK_512) {
+  LOG_DEBUG("Task 1 initialized!\n");
+  StatusCode ret;
+  while (true) {
+    for (size_t i = 0; i < LIST_SIZE; ++i) {
+      char *to_send = s_list[i];
```
mitchellostler, line 33: You can just send directly from the list, by passing s_list[i] as the argument to the below function.

projects/fw103/src/main.c
```diff
@@ -0,0 +1,72 @@
...
+        LOG_DEBUG("Sent \"%s\"\n", to_send);
+      }
+    }
+  }
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  // const char outstr[ITEM_SZ];
+  char *outstr;
```
mitchellostler, line 48: You can't use an empty pointer here. You need to pass it actually memory for the receive to read into, which is why the buffer of length ITEM_SZ was provided. Here it will try to write to the pointer outstr, which does not point to a valid region of memory, and will most likely cause a segfault. 
Let me know if you still have questions about the difference between the two.

projects/fw103/src/main.c
```diff
@@ -0,0 +1,72 @@
...
+  StatusCode ret;
+  while (true) {
+    ret = queue_receive(&s_queue1, &outstr, 100);
```
mitchellostler, line 51: Here you are passing the pointer to outstr, which is a char **. This is not what we want, outstr should just be passed as is

mitchellostler (commented):
Good job getting this done. The first task looks good, but the second one has some issues. If you can fix them up, and push your changes I can approve your FW 103

PR #56, opened 2022-05-30
FW103 Homework - Lauren Brecht

laure-nn:


projects/fw103/src/main.c
```diff
@@ -0,0 +1,69 @@
...
+      if (ret != STATUS_CODE_OK) {
+        LOG_DEBUG("Write to queue failed");
+      }
+    }
+  }
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  char outstr[5];
```
mitchellostler, line 45: Just a note, this should have been 6, to hold the string with the NULL character as well, but that's on me. Just thought I'd mention it.

projects/fw103/src/tasks/tasks.c
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+TASK(task1, TASK_STACK_512) {
+  int counter1 = 0;
+  while (true) {
+    LOG_DEBUG("task1 %d\n", counter1);
+    ++counter1;
+
+    prv_delay(pdMS_TO_TICKS(1000));
```
mitchellostler, line 25: Just a nitpick, but this won't actually convert to milliseconds. Since pdMS_TO_TICKS is called in prv_delay, it will convert the value twice.

mitchellostler (commented):
Overall looks good! You're good to go for 103.  I still haven't received your fw 102 yet, so if you could send that to me, that would be great thanks

PR #58, opened 2022-06-14
Run all test in task

ShiCheng-Lu:


ShiCheng-Lu:
there is problems with this.

libraries/ms-common/src/log.c
```diff
@@ -16,7 +17,9 @@ QueueHandle_t g_log_queue = NULL;
 
 void log_init(void) {
   g_log_queue = xQueueCreateStatic(QUEUE_LENGTH, ITEM_SIZE, s_log_array, &s_log_queue);
-  tasks_init_task(log_task, TASK_PRIORITY(tskIDLE_PRIORITY + 1), NULL);
+  if (log_task->context == NULL) {
+    tasks_init_task(log_task, TASK_PRIORITY(tskIDLE_PRIORITY + 1), (int *)1);
```
mitchellostler, line 21: Can we change this to checking if the task->handle is null?
Not a fan of using the context with an invalid pointer value, doesn't seem like great practice

Also, is there a reason you don't want to initialize the task twice for this and Adc?
I thought we confirmed that reinitialization wasn't going to be an issue

libraries/ms-common/inc/task_test_helpers.h
```diff
@@ -31,29 +31,10 @@
...
-
-// Implementation helper for TASK_TEST(), do not call directly.
-#define _TASK_TEST_IMPL(test_name)                        \
-  void test_##test_name(void) {                           \
-    tasks_init_task(test_name, TASK_TEST_PRIORITY, NULL); \
-    tasks_start();                                        \
-  }
+#ifndef TEST_PRE_TASK
+// define in-task/pre-task tests
+#define TEST_PRE_TASK
```
mitchellostler, line 38: Can you add a comment line or two here about what the specifics are for how to use these two

libraries/unity/auto/template_example.c
```diff
@@ -0,0 +1,84 @@
+/* AUTOGENERATED FILE. DO NOT EDIT. */
```
mitchellostler, line 1: Do we want this committed to master, or is this just a demo for the PR?

mitchellostler (commented):
Hey, looks great, just 3 comments. Also noticed you've got a merge conflict (I reduced the log size defines on my fsm PR, if you just switch it over to the other values, then it should be good to go).

mitchellostler (approved):

PR #59, opened 2022-06-14
Fwxv 95 create end task semaphore

awaterston01:
Created new method tasks_init. Initializes the s_end_task_sem and returns OK if initialized, uninitialized if not.

### DIFF
--- a/libraries/ms-common/inc/tasks.h
+++ b/libraries/ms-common/inc/tasks.h
@@ -87,5 +87,6 @@ StatusCode tasks_init_task(Task *task, TaskPriority priority, void *context);
 // This function should not return!
 void tasks_start(void);
 
-// Initialize the end task semaphore.
+// Initialize the task module. Must be called before tasks are initialized or the scheduler is
+// started.
 StatusCode tasks_init(void);
--- a/libraries/ms-common/src/tasks.c
+++ b/libraries/ms-common/src/tasks.c
@@ -65,8 +65,7 @@ void tasks_start(void) {
 }
 
 StatusCode tasks_init(void) {
-  // Initialize the task module. Must be called before tasks are initialized or the scheduler is
-  // started.
+  // Initialize the end task semaphore.
   s_end_task_handle = xSemaphoreCreateCountingStatic(MAX_NUM_TASKS, 0, &s_end_task_sem);
 
   if (s_end_task_handle == NULL) {


libraries/ms-common/src/tasks.c
```diff
@@ -59,3 +63,14 @@ void tasks_start(void) {
   LOG_CRITICAL("CRITICAL: scheduler stopped!\n");
 #endif
 }
+
+StatusCode tasks_init(void) {
+  // Initialize end task semaphore.
+  s_end_task_handle = xSemaphoreCreateCountingStatic(MAX_COUNT, INITIAL_COUNT, &s_end_task_sem);
```
mitchellostler, line 69: Like I mentioned above, just pass 0 here for initial count, and we will use MAX_NUM_TASKS for our max count

mitchellostler (commented):
Hey, Great work on this! A couple of nitpicks, if you can resolve them and put them up, that would be great.

mitchellostler:
Also, if you can assign me/jarvis as a reviewer next time, we'll get an email notif that your PR is up 😄

libraries/ms-common/inc/tasks.h
```diff
@@ -60,6 +61,13 @@
 // A convenience macro for making declaring priorities more readable.
 #define TASK_PRIORITY(prio) ((TaskPriority)prio)
 
+// Parameters that will be used to initialize the end semaphore.
+#define MAX_NUM_TASKS 15
+
+// Semaphore variables
+extern StaticSemaphore_t s_end_task_sem;
```
mitchellostler, line 68: Sorry, I missed this before, but these extern declarations shouldn't be here. Both these vars should only be declared statically in the .c file

libraries/ms-common/src/tasks.c
```diff
@@ -6,6 +6,10 @@
 #include "log.h"
 #include "status.h"
 
+// End task semaphore.
+StaticSemaphore_t s_end_task_sem;
```
mitchellostler, line 10: These declarations should be static, since we only want to access them from this file

libraries/ms-common/inc/tasks.h
```diff
@@ -82,3 +90,6 @@ StatusCode tasks_init_task(Task *task, TaskPriority priority, void *context);
 // Start the FreeRTOS scheduler to run the tasks that were previously initialized.
 // This function should not return!
 void tasks_start(void);
+
+// Initialize the end task semaphore.
```
mitchellostler, line 94: Can you change this to "Initialize the task module. Must be called before tasks are initialized or the scheduler is started"

mitchellostler (changes requested):
Sorry, realize I missed a couple things in the last one, just two more fixes

mitchellostler (approved):

PR #62, opened 2022-06-25
Add mutexes to SPI and I2C

Bafran:


mitchellostler (changes requested):
Hey, you've got the right idea but there's some issues with how it's currently set up. 
Basically, once we've acquired a mutex we can't return until we unlock the mutex. This means that we can't use the status_ok_or_return, or any macro that returns inside the mutex "critical  section" (from when it's locked to when it's unlocked).

To resolve this, just collect the return code, execute what's needed/not needed based on it's value, and then unlock the mutex and return the value. We basically just want to have a single exit point in any function, at the end. 
For any macros that have a return, just make them real functions with a StatusCode return type.

The comments I've listed here pretty go for both i2c and SPI

libraries/ms-common/src/arm/i2c.c
```diff
@@ -133,13 +141,16 @@ StatusCode i2c_read(I2CPort i2c, I2CAddress addr, uint8_t *rx_data, size_t rx_le
   }
   // CRITICAL_SECTION_AUTOEND;
 
-  I2C_TIMEOUT_WHILE_FLAG(i2c, I2C_FLAG_BUSY, SET);
+  status_ok_or_return(mutex_lock(&i2c_mutex, I2C_MUTEX_WAIT_MS));
 
-  status_ok_or_return(prv_transfer(i2c, addr, true, rx_data, rx_len, I2C_AutoEnd_Mode));
+  // Proceed if mutex lock was successful
+  StatusCode ret = STATUS_CODE_OK;
+  if (ret == STATUS_CODE_OK) ret = prv_i2c_timeout_while_flag(i2c, I2C_FLAG_BUSY, SET);
```
mitchellostler, line 148: Nit: Do these in bracket enclosed if statements. Single line if statements introduce the possibility for more errors

libraries/ms-common/inc/i2c.h
```diff
@@ -10,6 +10,8 @@
 #include "i2c_mcu.h"
 #include "status.h"
 
+#define I2C_MUTEX_WAIT_MS 0
```
mitchellostler, line 13: Can you make both these timeouts 100? They'll get changed later but this will make them usable for now

libraries/ms-common/src/arm/i2c.c
```diff
@@ -8,35 +8,21 @@
...
-    I2C_ClearFlag(s_port[i2c_port].base, I2C_FLAG_STOPF);    \
-  } while (0)
 
 typedef struct {
   uint32_t periph;
   I2C_TypeDef *base;
   I2CSettings settings;
 } I2CPortData;
 
+static Mutex i2c_mutex;
```
mitchellostler, line 24: Nit: can you prefix with s_ (since it's a static var)

libraries/ms-common/src/arm/i2c.c
```diff
@@ -79,16 +85,16 @@ static StatusCode prv_transfer(I2CPort port, uint8_t addr, bool read, uint8_t *d
 
   if (read) {
     for (size_t i = 0; i < len; i++) {
-      I2C_TIMEOUT_WHILE_FLAG(port, I2C_FLAG_RXNE, RESET);
+      prv_i2c_timeout_while_flag(port, I2C_FLAG_RXNE, RESET);
```
mitchellostler, line 88: You should status_ok_or_return these calls, since we want to know if these have failed

libraries/ms-common/src/arm/spi.c
```diff
@@ -16,6 +18,8 @@ static SpiPortData s_port[NUM_SPI_PORTS] = {
   [SPI_PORT_2] = { .rcc_cmd = RCC_APB1PeriphClockCmd, .periph = RCC_APB1Periph_SPI2, .base = SPI2 },
 };
 
+static Mutex spi_mutex;
```
mitchellostler, line 21: s_ prefix

mitchellostler (changes requested):
Just some last nitpicks, otherwise looks good to go

mitchellostler (approved):

PR #63, opened 2022-06-25
FWXV 2: Freertos FSM library

mitchellostler:
Still got some cleanup to do, and want to expand the demo project a bit, but I thought it should go up now that the structure is done and I have a fully functional demo project

### DIFF
--- a/libraries/ms-common/inc/fsm.h
+++ b/libraries/ms-common/inc/fsm.h
@@ -1,0 +1,79 @@
+#pragma once
+// Fsm.h
+// Module used to implement finite-state-machine behaviour
+// FSMs are automatically created with their own task which can be referenced
+// by the name passed to DECLARE_FSM(). 
+// 
+#include "fsm_impl.h"
+#include "status.h"
+#include "tasks.h"
+#include "notify.h"
+#include "semphr.h"
+
+#define FSM_PRIORITY 1
+#define FSM_TASK_STACK TASK_STACK_1024 // TODO: make user defined
+#define FSM_TIMEOUT_MS 100
+#define CYCLE_RX_MAX 15 // TODO: make this defined as part of project API
+
+struct Fsm;
+typedef uint8_t StateId;
+
+typedef void (*StateInputFunc)(struct Fsm *fsm, void *context);
+typedef void (*StateOutputFunc)(void *context);
+
+typedef struct FsmState {
+  StateId id;              // Unique State ID - Should be based on enumerated type
+  StateInputFunc inputs;   // Function used to parse inputs and transition if needed
+  StateOutputFunc outputs; // Output function associated with state
+} FsmState;
+
+typedef struct Fsm {
+  FsmState *curr_state;
+  const uint8_t num_states;
+  FsmState **transition_table;
+  SemaphoreHandle_t fsm_sem;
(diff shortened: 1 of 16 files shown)


mitchellostler:
Size info for the fsm demo program for your viewing pleasure 
  17612     956   10240   28808    7088 build/arm/bin/projects/fsm_demo

For some reason, the task stacks are showing up as  as 2048 bytes, even though they are initialized as 512. We will probably need to look into this at some point. It might just be that task stack size is in words 🤷 .

libraries/ms-common/inc/fsm.h
```diff
@@ -0,0 +1,139 @@
...
+
+// Defines a transition from StateId "from" -> StateId "to"
+// Must be declared in a transition list
+#define TRANSITION(from_state, to_state) _TRANSITION(from_state, to_state)
+
+#define fsm_init(fsm, settings, context) _fsm_init(fsm, settings, context)
+
+// Initiates a transition from the current state
+// Transition must exist in transition table
+StatusCode fsm_transition(Fsm *fsm, StateId to);
```
JarvisWeng, line 128: This will allow external tasks to forceably transition an FSM's state. Do we want to do that, or just let the FSM have total control over its own state?

libraries/ms-common/test/test_fsm.c
```diff
@@ -0,0 +1,137 @@
...
+  TEST_ASSERT_EQUAL(TEST_STATE_1, test1_fsm->curr_state->id);
+  // 1 -> 0
+  TEST_ASSERT_OK(fsm_transition(test1_fsm, TEST_STATE_0));
+  TEST_ASSERT_EQUAL(TEST_STATE_0, test1_fsm->curr_state->id);
+  // 0 -> 2
+  TEST_ASSERT_OK(fsm_transition(test1_fsm, TEST_STATE_2));
+  TEST_ASSERT_EQUAL(TEST_STATE_2, test1_fsm->curr_state->id);
+
+  // Re-initialize in state 1
+  prv_init_fsm(TEST_STATE_1);
```
JarvisWeng, line 113: So we can reinitialize an FSM multiple times?

JarvisWeng (approved):

libraries/ms-common/inc/fsm.h
mitchellostler: It is possible that this can be accessed externally, but that is not how it should be used. It isn't part of the main interface, since the only method that receives an fsm object as a parameter is the inputs function, but it could theoretically be accessed if the fsm_task->context is passed. 
I think we can leave it as is, and just enforce that only an FSM should change its own state via its input functions

mitchellostler: Alternatively, this may be useful if we need to forcibly transition back to an initial state from the error handler, but we could also do this through the regular interface with a notification, and program a check for it in each of the input functions

libraries/ms-common/test/test_fsm.c
mitchellostler: Theoretically yes, we could re-initialize to set a new initial state, and change the transition list/state list, but there wouldn't be a situation I can think of where we want to since all of those are statically declared and immutable. The initial state might be a good one I guess

PR #64, opened 2022-06-28
Rename mutex

jeyoung78:


libraries/ms-common/inc/mutex.h
```diff
@@ -30,3 +32,15 @@ StatusCode mutex_lock(Mutex *mutex, uint16_t ms_to_wait);
 // higher_priority_task_woken informs the caller whether a higher priority task has become unblocked
 // Returns STATUS_CODE_OK on success, STATUS_CODE_INTERNAL_ERROR on failure
 StatusCode mutex_unlock(Mutex *mutex);
+
+// wrapper over xSemaphoreCreateCountingStatic
```
mitchellostler, line 36: Change this comment to // Initialize counting semaphore with max and initial count

libraries/ms-common/inc/mutex.h
```diff
@@ -30,3 +32,15 @@ StatusCode mutex_lock(Mutex *mutex, uint16_t ms_to_wait);
...
+StatusCode sem_init(UBaseType_t uxMaxCount, UBaseType_t uxInitialCount, Semaphore *sem);
```
mitchellostler, line 37: The sem parameter should be first (just convention, and this should take a `uint32_t max_count` and a `uint32_t init_count` instead of what you have here. We're trying to use our own coding conventions and move away from the FreeRTOS standards. You will still need to pass UBaseType_t values to the FreeRTOS API functions, but you should do that in the functions themselves

libraries/ms-common/inc/mutex.h
```diff
@@ -30,3 +32,15 @@ StatusCode mutex_lock(Mutex *mutex, uint16_t ms_to_wait);
...
+
+// wrapper over xSemaphoreTake
```
mitchellostler, line 39: Same for these functions, the comments should be what they are actually doing, instead of what they are wrapping over. Let me know if you're not sure for any of them what the description should be, but see if you can figure it out

libraries/ms-common/src/mutex.c
```diff
@@ -35,3 +36,38 @@ StatusCode mutex_unlock(Mutex *mutex) {
...
+  if (sem->handle == NULL) {  // pxSemaphore is NULL
+    return STATUS_CODE_UNINITIALIZED;
+  } else {  // Semaphore is created succesfully
+    return STATUS_CODE_OK;
+  }
+}
+
+StatusCode sem_wait(Semaphore *sem, uint32_t timeout_ms) {
+  if (xSemaphoreTake(sem->handle, timeout_ms) ==
+      pdFALSE) {  // xTicksToWait expired without the semaphore becoming available
```
mitchellostler, line 51: Nit: Move the comment to a separate line above so the line doesn't wrap around

libraries/ms-common/src/mutex.c
```diff
@@ -35,3 +36,38 @@ StatusCode mutex_unlock(Mutex *mutex) {
...
+}
+
+StatusCode sem_post(Semaphore *sem) {
+  if (xSemaphoreGive(sem->handle) == pdFALSE) {  // an error occurred
+    return STATUS_CODE_INTERNAL_ERROR;
+  }
+  return STATUS_CODE_OK;  // Semaphore was released
+}
+
+StatusCode sem_num_items(Semaphore *sem) {
```
mitchellostler, line 64: Sorry, I may have messed this up in the description, but this should just return a uint32_t with the number of items returned from uxSemaphoreGetCount, it doesn't need to have any status code logic

projects/freertos/README.md
```diff
@@ -0,0 +1,16 @@
+<!--
```
mitchellostler, line 1: This file should get removed, and all the other ones below that are not part of what you are working on. I think they're just left over from the 103 hw

libraries/ms-common/src/mutex.c
```diff
@@ -35,3 +36,38 @@ StatusCode mutex_unlock(Mutex *mutex) {
   }
   return STATUS_CODE_OK;
 }
+
+StatusCode sem_init(UBaseType_t uxMaxCount, UBaseType_t uxInitialCount, Semaphore *sem) {
+  sem->handle = xSemaphoreCreateCountingStatic(uxMaxCount, uxInitialCount, sem->buffer);
```
mitchellostler, line 41: This method expects a pointer for it's third argument (ie &sem->buffer instead of sem->buffer)

mitchellostler (changes requested):
Overall looks good! I added a few things that will need to get fixed, but you're close.

A couple notes:
- Your build is still failing, I added a comment that will help fix it. Let me know if you need me to show you how to run locally again
- Nitpick: In terms of naming your branches they should follow the form fwxv_<ticket#>_<description_of_what_you_are_working_on>. PR titles should look like FWXV <Ticket#>: Description
Makes it a lot easier for me to look through  (Also the angle brackets are placeholders, don't add them in lol)

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

libraries/ms-common/src/mutex.c
```diff
@@ -35,3 +36,31 @@ StatusCode mutex_unlock(Mutex *mutex) {
   }
   return STATUS_CODE_OK;
 }
+
+StatusCode sem_init(uint32_t uxMaxCount, uint32_t uxInitialCount, Semaphore *sem) {
```
mitchellostler, line 40: Should do a NULL check here:
`if (sem == NULL )` return invalid args. 
Should always do a null check since we don't want to dereference null pointers (auto crash). we don't need to do it for the buffer, since the FreeRTOS function does it, but we should do it for the pointer we dereference (use the `->` to get it's internal members).

Also Nit:
the Semaphore *sem parameter should be first in the function definition, just how we generally write things

libraries/ms-common/src/mutex.c
```diff
@@ -35,3 +36,31 @@ StatusCode mutex_unlock(Mutex *mutex) {
...
+  sem->handle = xSemaphoreCreateCountingStatic(uxMaxCount, uxInitialCount, &sem->buffer);
+  if (sem->handle == NULL) {
+    return STATUS_CODE_UNINITIALIZED;
+  } else {
+    return STATUS_CODE_OK;
+  }
+}
+
+StatusCode sem_wait(Semaphore *sem, uint32_t timeout_ms) {
```
mitchellostler, line 49: Should NULL check sem in each of these functions.

mitchellostler (changes requested):
Hey, just a couple of things, one nitpick and then adding a few null checks on inputs. Otherwise looks good!

mitchellostler:
Hey, you've got a merge conflict which you'll need to resolve. let me know if you need a hand with that

mitchellostler (approved):

PR #65, opened 2022-06-29
Update new_target.py to generate master_task [FWXV-94]

rbrtknwls:
My work for FWXV-94


scons/new_target.py
```diff
@@ -30,6 +30,74 @@
 
     """)
 
+
+MAIN_TEMPLATE = """\
```
JarvisWeng, line 34: I would put this in a different file so that's its easier to modify later. And then read in the file

scons/new_target.py
```diff
@@ -30,6 +30,74 @@
...
+    #include <stdio.h>
+
+    #include "can.h"
+    #include "can_board_ids.h"
+    #include "log.h"
+    #include "new_can_setters.h"
+    #include "tasks.h"
```
JarvisWeng, line 41: Only include `log.h` and `tasks.h`. The rest were specific for my example

scons/new_target.py
```diff
@@ -30,6 +30,74 @@
...
+
+    void wait_tasks(int t)
```
JarvisWeng, line 43: This has an actual implementation elsewhere so get rid of this function

scons/new_target.py
```diff
@@ -30,6 +30,74 @@
...
+         if (!(counter%100))
+            run_slow_counter();
+         vTaskDelay(pdMS_TO_TICKS(1000));
+         ++counter;
+       }
+    }
+
+    void run_fast_cycle()
+    {
+      run_critical_tasks();
```
JarvisWeng, line 64: Just keep all the `run..._cycle()` function empty

scons/new_target.py
```diff
@@ -30,6 +30,74 @@
...
+      wait_tasks(1);
+    }
+
+    void run_medium_cycle()
+    {
+      run_can_rx_cycle();
```
JarvisWeng, line 70: Make this empty

scons/new_target.py
rbrtknwls: That makes alot of sense, I made a .txt document within a new template folder

scons/new_target.py
rbrtknwls: Gotcha! Should be fixed now

rbrtknwls:
Thanks for reviewing my code Jarvis, I made the changes you recommended so I think I'm ready for another round of review!

JarvisWeng (approved):
Looks good, I'll merge

PR #66, opened 2022-06-29
FWXV-109, log on stack overflow

awaterston01:


libraries/ms-freertos/inc/FreeRTOSConfig.h
```diff
@@ -59,7 +59,7 @@ extern uint32_t SystemCoreClock;
 // Hook function related definitions
 #define configUSE_IDLE_HOOK 0
 #define configUSE_TICK_HOOK 0
-#define configCHECK_FOR_STACK_OVERFLOW 0
+#define configCHECK_FOR_STACK_OVERFLOW 2
```
awaterston01, line 62: for method 2 of stack overflow detection: https://www.freertos.org/Stacks-and-stack-overflow-checking.html

mitchellostler (approved):

PR #71, opened 2022-07-10
python script to generate memory info

harsh-patell:


mitchellostler (changes requested):
Overall, looks great! I really like how short and sweet your script is.
A couple things:
1. The script isn't running with the build as you have it setup right now. If you had already build the project, then it would be, but for a clean build (where the binaries haven't been built yet) it won't. I think it's just a matter of moving the if MEM_REPORT/call down to after the targets have been built (around line 311). You can verify that it is working by running scons clean first and then your mem cmd
2. The addoption/getoption logic should be moved to where the rest of the other options are for consistency

mitchellostler (approved):
Great work, I'm very happy with this!

PR #75, opened 2022-07-16
FW103 Homework

elbertchen1:
FW103 Homework by Elbert Chen


projects/FW_103/src/main.c
```diff
@@ -0,0 +1,90 @@
...
+
+TASK(task1, TASK_STACK_512) {
+  LOG_DEBUG("Task 1 initialized!\n");
+  StatusCode ret;
+  char word[ITEM_SZ];
+
+  while (true) {
+    for (int count = 0; count < QUEUE_LEN; count++)
+    {
+      for (int count1 = 0; count1 < ITEM_SZ - 1; count1++)
```
mitchellostler, line 43: For future reference, you don't need to do this. You can just pass the s_list index itself instead of copying the word over

projects/FW_103/src/main.c
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  const char outstr[ITEM_SZ];
+  StatusCode ret;
+  while (true) {
+    ret = queue_receive(&s_queue1, &s_queue1_buf, 1000);
```
mitchellostler, line 65: You should not be receiving the data to here. This will break your queue, as the array you are passing is already used as memory for the queue itself. You should be using the char array that is provided with the size of the word. (top of task)

mitchellostler (changes requested):
Hey Elbert, there is only one major issue with your implementation, but otherwise looks good. Were you able to run the code for the queues task? 

Anyways, if you could fix it and push your changes that would be great

mitchellostler (changes requested):
Hey Elbert, there is only one major issue with your implementation, but otherwise looks good. Were you able to run the code for the queues task? 

Anyways, if you could fix it and push your changes that would be great

elbertchen1:
Hey Mitchell, I made the changes you requested. Thanks!

mitchellostler (commented):
Looks good!

PR #76, opened 2022-07-16
Add extended ids

Bafran:


can/src/can.c
```diff
@@ -146,8 +146,8 @@ StatusCode can_transmit(const CanMessage *msg, const CanAckRequest *ack_request)
 
   CanId can_id = { .raw = msg->msg_id };
 
-  // TODO: make CAN with extended
-  return can_hw_transmit(can_id.raw, false, msg->data_u8, msg->dlc);
+  bool extended = (can_id.raw >= CAN_MSG_ID_STD_MAX);
+  return can_hw_transmit(can_id.raw, extended, msg->data_u8, msg->dlc);
```
JarvisWeng, line 150: Does the `STM` version need any configuration for extended?

can/inc/can_msg.h
```diff
@@ -9,7 +9,8 @@
 #define CAN_MSG_INVALID_DEVICE (UINT16_MAX)
 
 #define CAN_MSG_MAX_DEVICES (1 << 4)
-#define CAN_MSG_MAX_IDS (1 << 6)
+#define CAN_MSG_MAX_IDS (1 << 24)
+#define CAN_MSG_ID_STD_MAX (1 << 11)
```
JarvisWeng, line 13: You'll also need to change this
https://github.com/uw-midsun/fwxv/blob/main/libraries/codegen/templates/can_board_ids.h.jinja

To keep it simple, we can just make every message an extended message.
So we will want the first 4 bits to indicate the `board_id` and the other bits to indicate the 'message' id

JarvisWeng: Maybe try to change it to a macro to make it simpler

JarvisWeng (approved):

PR #79, opened 2022-07-24
Fwxv 83 add can oneshot mode

123Logan321:


SConstruct
```diff
@@ -188,9 +188,10 @@ def generate_can_files(env, target=[], source=[], project=PROJECT):
     # TODO: Fix up system_can.dbc output directory
     for template in templates:
         if "can_board_ids" in template.name:
+            # TODO: Need to fix for multiple projects that depend on can_codegen.h
             env.Command(
                 LIBRARIES_INC_DIR.File(template.name[:-6]),
-                source,
+                [], # TODO: Fix this part
```
JarvisWeng, line 194: What happened here?

can/src/can.c
```diff
@@ -97,13 +97,21 @@ StatusCode can_init(CanStorage *storage, const CanSettings *settings)
   // Initialize hardware settings
   status_ok_or_return(can_hw_init(&s_can_storage->rx_queue, settings));
 
-  // Create RX and TX Tasks
-  // TODO: Figure out priorities
-  status_ok_or_return(tasks_init_task(CAN_RX, TASK_PRIORITY(2), NULL));
-  status_ok_or_return(tasks_init_task(CAN_TX, TASK_PRIORITY(2), NULL));
+  // Disable tasks when in can_one_shot_mode 
+  if (settings->mode == 1){
```
JarvisWeng, line 101: Remove please

can/src/can.c
```diff
@@ -150,7 +158,21 @@ StatusCode can_transmit(const CanMessage *msg, const CanAckRequest *ack_request)
   return can_hw_transmit(can_id.raw, false, msg->data_u8, msg->dlc);
 }
 
-StatusCode can_add_filter(CanMessageId msg_id) {
+StatusCode can_add_filter_in(CanMessageId msg_id) {
+  if (s_can_storage == NULL) {
```
JarvisWeng, line 162: You should also have a global variable to disable `filter_in` or `filter_out`. Because we only want to allow one of these functions to be used in a project. Not both at the same time.

JarvisWeng: Also have you test and verify that it's working?

libraries/codegen/boards/boards.yaml
```diff
@@ -23,4 +23,5 @@
     - steering
 
     - new_can
+    - can_one_shot_mode
```
JarvisWeng, line 26: Yeah, this is your SCons issue that you will need to fix. `one shot` mode projects shouldn't need to generate a yaml or things

SConstruct
123Logan321: oh I think you left it here as a reminder for something when you were helping me with the ticket

can/src/can.c
123Logan321: Yep I tested it with vcan and it works

can/src/can.c
123Logan321: Do you mean remove the whole conditional statement or just the comment?

SConstruct
JarvisWeng: Oh right, but that's for you to fix. Since you'll need to properly label the sources to build the `can` target. This has more to do with `Scons` and I can help you with that

can/src/can.c
JarvisWeng: The if statement since you don't need that LOG

PR #82, opened 2022-09-07
Fwxv 83 can one shot mode

123Logan321:


libraries/codegen/boards/can_one_shot_mode.yaml
```diff
@@ -0,0 +1,10 @@
+---
+  Messages:
+    bps_heartbeat:
```
JarvisWeng, line 3: Change this to `one_shot_msg`

libraries/codegen/boards/can_one_shot_mode.yaml
```diff
@@ -0,0 +1,10 @@
+---
+  Messages:
+    bps_heartbeat:
+      id: 0
+      target:
+        - centre_console
```
JarvisWeng, line 6: Make target empty

PR #83, opened 2022-09-09
Fwxv 83can one shot mode rebased off of main

123Logan321:


can/src/can.c
```diff
@@ -19,6 +19,11 @@ static StaticSemaphore_t s_can_rx_sem;
 static SemaphoreHandle_t s_can_tx_sem_handle;
 static StaticSemaphore_t s_can_tx_sem;
 
+bool CAN_FILTER_IN_EN = true;
+
+#define CAN_RX_EVENT (1 << 0)
+#define CAN_FILTER_IN_EN 0
```
JarvisWeng, line 25: Get rid of these 2 #defines you don't need them

can/src/can.c
```diff
@@ -19,6 +19,11 @@ static StaticSemaphore_t s_can_rx_sem;
 static SemaphoreHandle_t s_can_tx_sem_handle;
 static StaticSemaphore_t s_can_tx_sem;
 
+bool CAN_FILTER_IN_EN = true;
```
JarvisWeng, line 22: make this static and `s_can_filter_in_en`. Also this should have 3 different values not 2.

can/src/can.c
```diff
@@ -19,6 +19,9 @@ static StaticSemaphore_t s_can_rx_sem;
 static SemaphoreHandle_t s_can_tx_sem_handle;
 static StaticSemaphore_t s_can_tx_sem;
 
+//takes 1 for filter_in, 2 for filter_out and default is unset
+static int s_can_filter_in_en;
```
JarvisWeng, line 23: You should keep this with a default value, it should set it to be initialized with 0

JarvisWeng (approved):

PR #84, opened 2022-09-10
Change queues to UART

Bafran:


libraries/ms-common/inc/log.h
```diff
@@ -4,9 +4,10 @@
 #include <stdlib.h>
 
 #include "FreeRTOS.h"
-#include "queue.h"
+#include "retarget_cfg.h"
```
mitchellostler, line 7: You shouldn't need retarget, this is what we're trying to do away with

libraries/ms-common/inc/log.h
```diff
@@ -35,38 +37,52 @@ void log_init(void);
...
-    if (xTaskGetSchedulerState() != taskSCHEDULER_RUNNING) {                            \
-      printf("%s", log_buffer);                                                         \
-    } else {                                                                            \
-      if (g_log_queue == NULL) {                                                        \
-        printf("Error: Logs init not initialized \n");                                  \
-        vTaskEndScheduler();                                                            \
-      }                                                                                 \
-      xQueueSendToBack(g_log_queue, log_buffer, 0);                                     \
-    }                                                                                   \
+#ifdef x86
```
mitchellostler, line 40: You've got your X86 and arm versions backwards at the moment. The one with the printf should happen if x86 is defined (this means we are using x86) and we are able to print to the console.

mitchellostler (changes requested):
Ok, so there's quite a few changes that need to be made, partially because us having a queue-based UART module makes more things possible, partially because I am seeing better ways to go about this. 

First steps:
- Right now the versions of logging you have for x86/arm are backwards
- The x86 should just be calling printf with the arguments that are currently supplied to snprintf, that's it. so replace snprintf with printf, and remove everything else. (We will probably want to go through the queue once we have an x86 version of Uart, but for now keep it simple. )

The arm version is a bit more fun, so here's what we're going to do:
- get rid of the entire log.c file. We're not going to be using a task for logging anymore, so we can nix it all.  
- Create an extern log buffer array of size max_log_size, and an extern const Mutex pointer
- Create a `log_init()` macro, which will provide the implementation of the log buffer, and the mutex, and initialize them. 
- Each log call will write to the same log buffer, but must lock the mutex before writing to the log buffer and sending to Uart, and then unlock it after uart has written to the queue

The last part is definitely a bit more complex so let me know if you need me to explain it in more detail in a meeting or something

libraries/ms-common/inc/log.h
```diff
@@ -15,58 +16,51 @@ DECLARE_TASK(log_task);
...
-    } else {                                                                            \
-      if (g_log_queue == NULL) {                                                        \
-        printf("Error: Logs init not initialized \n");                                  \
-        vTaskEndScheduler();                                                            \
-      }                                                                                 \
-      xQueueSendToBack(g_log_queue, log_buffer, 0);                                     \
-    }                                                                                   \
+#define log_init()                        \
+  do {                                    \
+    extern char log_buffer[MAX_LOG_SIZE]; \
```
mitchellostler, line 43: The extern char log buffer should be renamed to g_log_buffer, and the extern should be outside of the macro, basically where `static char log_buffer` is right now. That is the reference we are creating so that no matter where the macro is called, it will find that buffer as long as log.h is included. 

Where you have extern char log_buffer right now should be changed to just `char g_log_buffer[MAX_LOG_SIZE]`. This is where the actual memory will exist. since log init will only be called once (in main) we will have one buffer that we can use for our logs.

libraries/ms-common/inc/log.h
```diff
@@ -15,58 +16,51 @@ DECLARE_TASK(log_task);
...
-      if (IN_ORDER_LOGS) {                                                              \
-        xQueueSendToBack(g_log_queue, log_buffer, 0);                                   \
-      } else {                                                                          \
-        xQueueSendToFront(g_log_queue, log_buffer, 0);                                  \
-      }                                                                                 \
-    }                                                                                   \
+#ifdef MS_PLATFORM_X86
+#define LOG(level, fmt, ...)                                                \
+  do {                                                                      \
+    printf("[%u] %s:%u: " fmt, (level), __FILE__, __LINE__, ##__VA_ARGS__); \
```
mitchellostler, line 51: This looks good 👍

libraries/ms-common/inc/log.h
```diff
@@ -15,58 +16,51 @@ DECLARE_TASK(log_task);
...
   } while (0)
+#else
+#define LOG(level, fmt, ...)                                                                  \
+  do {                                                                                        \
+    mutex_lock(&s_log_mutex, 0);                                                              \
```
mitchellostler, line 56: We can't have it be zero wait time, since we may have logs waiting that we don't want to lose. You should add a define for LOG_TIMEOUT_MS and make it like 10 rn. 

Additionally, you should make it so that the log functionality is executed **only** if the mutex_lock is successful.

libraries/ms-common/inc/log.h
```diff
@@ -15,58 +16,51 @@ DECLARE_TASK(log_task);
...
+    size_t msg_size = (size_t)snprintf(log_buffer, MAX_LOG_SIZE, "[%u] %s:%u: " fmt, (level), \
+                                       __FILE__, __LINE__, ##__VA_ARGS__);                    \
+    if (xTaskGetSchedulerState() != taskSCHEDULER_RUNNING) {                                  \
```
mitchellostler, line 59: I'm curious if this still works. If it does, it may make sense to continue using retarget with printf, just when the scheduler isn't started, as long as that doesn't break the functionality of using uart_tx for logs normally

mitchellostler (changes requested):
Generally looks good, just a couple comments.
It would also be great if you could test this on Hardware (and would be good experience) sometime if you have the chance. You can ask Jarvis for the intro, or I'm happy to walk you through the hw process on Zoom

mitchellostler (approved):

JarvisWeng:
😭 can we get this in?

mitchellostler:
@Bafran If you just rebase this I think it should resolve things? I was waiting to merge this in until uart is tested, but I think we should just go ahead and do it

mitchellostler (approved):

PR #86, opened 2022-09-15
Fwxv 102 create setter jinjas to populate setter functions

laure-nn:
test_generator.py seems to ignore the contents of my yaml file when I try running it.

### DIFF
--- a/libraries/test_gen/src/setter_template.jinja
+++ b/libraries/test_gen/src/setter_template.jinja
@@ -1,11 +1,9 @@
+{% set variables = data["variables"] -%}
 #ifdef MS_TEST
-
 {% for var, type in variables.items() %}
 void set_{{var}}({{type}} val)
 {
     {{var}} = val;
 }
-
 {% endfor %}
-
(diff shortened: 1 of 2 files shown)


libraries/test_gen/src/setter_template.jinja
```diff
@@ -0,0 +1,11 @@
+#ifdef MS_TEST
+
```
JarvisWeng, line 2: Oh for this simply add this line
```jinja
{% set variables = data["variables"] -%}
```

All data from the `yaml` is stored in a jinja variable called `data`

libraries/test_gen/src/setter_template.jinja
```diff
@@ -1,11 +1,9 @@
+{% set variables = data["variables"] -%}
```
JarvisWeng, line 1: I think this file should be named `setter_template.c.jinja`?
Do you see the `.c` when it outputs a file?

JarvisWeng (approved):

PR #88, opened 2022-09-21
Integrate F1 build files into system, refactor peripheral drivers

mitchellostler:
This seems like a HUGE PR but don't be daunted. Mostly just  a lot of file deletions/additions for the build system. 

- replaced stm32f0xx with stm32f10x everywhere
- Replaced all of STDPeriph/CMSIS files with new versions for F1

- Refactored adc.c/gpio.c/gpio_it.c/stm32f10x_interrupts.c. There are really the only ones that need review


libraries/ms-common/src/arm/adc.c
```diff
@@ -8,30 +8,22 @@
 #include "log.h"
 #include "semphr.h"
 #include "status.h"
-#include "stm32f0xx.h"
+#include "stm32f10x.h"
 
-enum {
```
JarvisWeng, line 13: Is there a reason to get rid of these because I see you substitute it a lot later

libraries/ms-common/src/arm/gpio.c
```diff
@@ -5,69 +5,43 @@
...
+  // Create init struct for stdperiph call to initialize pin
+  GPIO_InitTypeDef init_struct = {
+    .GPIO_Pin = pin,
+    .GPIO_Speed = GPIO_Speed_50MHz,  // Default to fastes speed
+    .GPIO_Mode = pin_mode,
+  };
 
   // Set the pin state.
-  gpio_set_state(address, settings->state);
+  gpio_set_state(address, init_state);
```
mitchellostler, line 44: I'm probably just going to remove the init state all together

libraries/ms-common/src/arm/gpio_it.c
```diff
@@ -55,14 +45,8 @@ StatusCode gpio_it_register_interrupt(const GpioAddress *address, const Interrup
...
-  // Try to register on NVIC and EXTI. Both must succeed for the callback to be
-  // set.
-  SYSCFG_EXTILineConfig(address->port, address->pin);
-
-  status_ok_or_return(stm32f0xx_interrupt_exti_enable(address->pin, settings));
-
-  uint8_t irq_channel = prv_get_irq_channel(address->pin);
-  status_ok_or_return(stm32f0xx_interrupt_nvic_enable(irq_channel, settings->priority));
+  // Register exti channel and enable interrupt
+  status_ok_or_return(stm32f10x_interrupt_exti_enable(address, settings));
```
mitchellostler, line 49: Exti functionality moved to stm32f10x_interrupt.h

libraries/ms-common/src/arm/i2c.c
```diff
@@ -6,26 +6,50 @@
...
+typedef enum I2CMode {
+  I2C_MODE_TRANSMIT = 0,
+  I2C_MODE_RECEIVE,
+  NUM_I2C_MODES,
+} I2CMode;
+
+// Used to protect I2C resources
+// All data tx'd and rx'd through queue
+// Transactions on each I2C port protected by Mutex
+typedef struct {
```
mitchellostler, line 26: Use one queue for tx/rx since only one can happen at a given time

mitchellostler: Queue protected by mutex, so will never have multiple transactions using it at a time

libraries/ms-common/src/arm/i2c.c
```diff
@@ -6,26 +6,50 @@
 #include "i2c.h"
 
 #include <stdbool.h>
```
mitchellostler, line 8: SPI and I2C both use a method very similar to UART. 

Basically, calling a method  (read/write) will enable interrupts for that type of transaction, and then all data will go through a queue. 

For read, the ISR triggered by a read condition puts data on the queue, up to rx_len

For write, all data is put onto the queue initially, and then each sequential TXE (tx register empty) will put the next byte of data on the line

libraries/ms-common/src/arm/adc.c
mitchellostler: The new symbols for this that I use have different values, and I don't really want to convert between the enum and the real value each time I use the value.

mitchellostler:
This Commit [92f338a](https://github.com/uw-midsun/fwxv/pull/88/commits/92f338ad82cbfd3cfa728f96a31e185224e0e9ac)
is the one that has all the real changes needed for review. I will be continually updating as I see things that need to change

JarvisWeng:
> This Commit [92f338a](https://github.com/uw-midsun/fwxv/pull/88/commits/92f338ad82cbfd3cfa728f96a31e185224e0e9ac) is the one that has all the real changes needed for review. I will be continually updating as I see things that need to change

Lol this doesn't exist anymore

libraries/ms-common/src/arm/adc.c
JarvisWeng: I see

libraries/ms-common/src/arm/adc.c
```diff
@@ -116,151 +114,177 @@ static uint16_t prv_get_vdda(uint16_t reading) {
...
-static StatusCode prv_check_channel_enabled(uint8_t adc_channel) {
-  if (!(ADC1->CHSELR & ((uint32_t)1 << adc_channel))) {
-    return status_code(STATUS_CODE_EMPTY);
+// Checks if a channel has been registered
+static StatusCode prv_check_channel_enabled(uint8_t channel) {
+  bool found = false;
+  // Check each index in the sequence to see if it has a matching channel
+  for (uint8_t index = 0; index < s_adc_status.active_channels; index++) {
+    if (s_adc_sequence[index] == &s_adc_stores[channel]) {
+      found = true;
```
JarvisWeng, line 123: maybe add a break here

libraries/ms-common/src/arm/adc.c
```diff
@@ -116,151 +114,177 @@ static uint16_t prv_get_vdda(uint16_t reading) {
...
-        // YIELD_FROM_ISR should be at the end of the function
-        // even though it has same behaviour on stm32 as using notify_from_isr
-        xTaskNotifyFromISR(s_adc_stores[channel].task->handle, 1u << s_adc_stores[channel].event,
-                           eSetBits, &xHigherPriorityTaskWoken);
+    s_adc_sequence[s_adc_status.sequence]->reading = ADC_GetConversionValue(ADC1);
+    if (s_adc_sequence[s_adc_status.sequence]->task != NULL) {
+      // cannot use notify here because we need additional processing after notify
+      // YIELD_FROM_ISR should be at the end of the function
+      // even though it has same behaviour on stm32 as using notify_from_isr
+      xTaskNotifyFromISR(s_adc_sequence[s_adc_status.sequence]->task->handle,
```
JarvisWeng, line 259: So does any task that uses ADC need to register on this notify?

can/src/arm/can_hw.c
```diff
@@ -192,24 +189,46 @@ bool can_hw_receive(uint32_t *id, bool *extended, uint64_t *data, size_t *len) {
   return true;
 }
 
-void CEC_CAN_IRQHandler(void) {
+// TX handler
+void USB_HP_CAN1_TX_IRQHandler(void) {
```
JarvisWeng, line 193: I think Logan's PR is going to conflict with this, but we'll see who goes in first

JarvisWeng (commented):
Overall looks fine. Didn't go into the nitty gritty details though

libraries/ms-common/src/arm/adc.c
mitchellostler: Good point, I was going to leave this in, but it doesn't make sense to have. 
If we want an adc reading, we can either read the latest one from continuous, or trigger one and wait. 
There's not really a scenario when we need to get notified

can/src/arm/can_hw.c
```diff
@@ -256,7 +254,37 @@ void CEC_CAN_IRQHandler(void) {
   } else if (CAN_GetITStatus(CAN_HW_BASE, CAN_IT_ERR) == SET) {
     LOG_CRITICAL("Bus Unavailable");
   }
+  CAN_ClearITPendingBit(CAN_HW_BASE, CAN_IT_TME);
+}
+
+void USB_LP_CAN1_RX0_IRQHandler(void) {
```
JarvisWeng, line 260: Hm, it doesn't really show it, but I think you overwrote Logan's stuff

JarvisWeng (commented):
Also can you update the `SConstruct` in the `openocd` part, just change the `stm32f0....` to `...f1...`.

mitchellostler:
@JarvisWeng fixed CAN, I goofed on what the different ISRs were supposed to do. 
The other thing that I realized is that we won't be able to use LOG_DEBUG from any of the ISRs at the current moment. 
Once this gets merged, I will get Adel to add LOG_FROM_ISR to his Log PR and we can merge that in

JarvisWeng (dismissed):

SConstruct
```diff
@@ -248,9 +248,9 @@ if PLATFORM == 'arm' and TYPE == 'project':
             OPENOCD,
             '-s {}'.format(OPENOCD_SCRIPT_DIR),
             '-f interface/{}.cfg'.format(PROBE),
-            '-f target/stm32f0x.cfg',
-            '-f {}/stm32f0-openocd.cfg'.format(PLATFORM_DIR),
-            '-c "stm32f0x.cpu configure -rtos FreeRTOS"',
+            '-f target/stm32f1x.cfg',
+            '-f {}/stm32f1-openocd.cfg'.format(PLATFORM_DIR),
```
JarvisWeng, line 252: this should be `stm32f0-openocd.cfg` or you can change the file name in the `platform` directory

JarvisWeng:
Nvm, your branch doesn't work on the board

JarvisWeng (approved):
I'll just approve and test it once I have the time

PR #91, opened 2022-09-26
Create getter functions

A3sanjay:


JarvisWeng (commented):
Wait `_filter.h.jinja` should exist on `main`. You shouldn't delete it

libraries/codegen/boards/setter_example.yaml
```diff
@@ -0,0 +1,5 @@
+---
```
JarvisWeng, line 1: You shouldn't need this in `codegen` library

libraries/codegen/templates/_filters.h.jinja
```diff
@@ -11,6 +11,10 @@
 void add_all_can_filters()
 {
 {%- for message, message_data in messages.items() -%} 
+<<<<<<< HEAD
```
JarvisWeng, line 14: Looks like you rebase/merged incorrectly

libraries/codegen/templates/setter_template.h.jinja
```diff
@@ -0,0 +1,9 @@
+{% set variables = data["variables"] %}
```
JarvisWeng, line 1: You shouldn't need a `setter_template.h.jinja`. That's already created in `test_gen`

libraries/codegen/test_generator.py
```diff
@@ -0,0 +1,105 @@
+"""This function reads in yaml files, and populates jinja to make getter and setter functions"""
```
JarvisWeng, line 1: Don't need this since `test_generator.py` already exists

PR #93, opened 2022-09-30
Power Select Project Structure

A3sanjay:


JarvisWeng (commented):
Could you get rid of the `_tasks.c` file please? There is also an autogenerator for them so no need to make new files

JarvisWeng (approved):

PR #95, opened 2022-10-04
Fwxv 999 louis 103 hw

Scrubpai:
Completed FW103 HW Tasks


projects/fw103tasks/src/queues1.c
```diff
@@ -0,0 +1,67 @@
...
+        LOG_DEBUG("write to queue failed!\n");
+      }
+      delay_ms(100);
+    }
+  }
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  const char outstr[ITEM_SZ];
```
JarvisWeng, line 40: Technically this just overwrites the first entry over and over again. So the other spaces in the array are never used

JarvisWeng (commented):
Otherwise, it looks fine. Good job

PR #99, opened 2022-10-19
SPI x86 implementation

laure-nn:
Should the buffers be cleared after they're accessed in spi_rx and spi_get_tx?
Also, am I missing anything else?

### DIFF
--- a/libraries/ms-common/inc/spi.h
+++ b/libraries/ms-common/inc/spi.h
@@ -55,3 +55,11 @@ StatusCode spi_cs_get_state(SpiPort spi, GpioState *input_state);
 // RX.
 StatusCode spi_exchange(SpiPort spi, uint8_t *tx_data, size_t tx_len, uint8_t *rx_data,
                         size_t rx_len);
+
+#ifdef x86
+// Reads data from the tx buffer into data
+StatusCode spi_get_tx(uint8_t *data, uint8_t len);
+
+// Writes from data into the rx buffer
+StatusCode spi_set_rx(uint8_t *data, uint8_t len);
+#endif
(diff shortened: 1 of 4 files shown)


libraries/ms-common/inc/spi.h
```diff
@@ -55,3 +55,11 @@ StatusCode spi_cs_get_state(SpiPort spi, GpioState *input_state);
 // RX.
 StatusCode spi_exchange(SpiPort spi, uint8_t *tx_data, size_t tx_len, uint8_t *rx_data,
                         size_t rx_len);
+
```
mitchellostler, line 58: This looks good

libraries/ms-common/src/x86/spi.c
```diff
@@ -1,40 +1,72 @@
...
+
+  status_ok_or_return(mutex_init(&s_spi_mutex));
+
   return STATUS_CODE_OK;
 }
 
 StatusCode spi_tx(SpiPort spi, uint8_t *tx_data, size_t tx_len) {
-  return STATUS_CODE_OK;
+  if (mutex_lock(&s_spi_mutex, SPI_MUTEX_WAIT_MS) == STATUS_CODE_OK) {
+    if (tx_len > SPI_BUF_SIZE) {
```
mitchellostler, line 32: Error checking should generally be outside of mutex locks. You want to do all your checks at the beginning if you can. Also, if this were to return after locking, the mutex would never get unlocked (which is not good)

mitchellostler (changes requested):
Overall the approach is right! what you have here will work, aside from the few suggestions. 

The SPI library has changed since this ticket was assigned, and so if possible I'd like you to adjust what you've got here to use queues instead of the buffers you're using. 

If you take a look at [this file](https://github.com/uw-midsun/fwxv/blob/3fdf217e925cc2f92be0f234079ca5007156657f/libraries/ms-common/src/arm/spi.c), there's a SpiBuffer struct which would you can use, and it provides a tx and rx queue, along with a mutex for accessing them. Take a look at how the reading/writing to the queues works in this spi implementation here, and let me know if you have any questions. It's definitely a bit more tricky, but it will be a good thing to learn :)

mitchellostler:
Looks great! Unfortunately the build is broken because I added queue_reset() to my PR but it's not on master yet. If you want to copy it over to queue.h/c on your branch that would be a good idea, then we can go ahead and merge this

mitchellostler (approved):

PR #100, opened 2022-10-27
Scons add new task target back

ShiCheng-Lu:


SConstruct
```diff
@@ -170,6 +169,8 @@ SConscript('scons/test.scons', exports='VARS')
 # Helper targets
 ###########################################################
 SConscript('scons/new_target.scons', exports='VARS')
+SConscript('scons/new_task.scons', exports='VARS')
```
JarvisWeng, line 172: Can't you just add it as a target here instead of modifying my python file?

JarvisWeng (approved):

PR #102, opened 2022-10-29
mci fsm

devAdhiraj:
Added transition logic for mci fsm


projects/mci/src/mci_fsm.c
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void prv_mci_fsm_drive_input(Fsm *fsm, void *context) {
+  uint32_t notif;
+  notify_get(&notif);
+  if (notify_check_event(&notif, MCI_FSM_GOTO_OFF)) {
+    fsm_transition(fsm, MCI_FSM_STATE_OFF);
+  } else if (notify_check_event(&notif, MCI_FSM_GOTO_REVERSE)) {
+    if (s_storage.velocity == 0) {
+      fsm_transition(MCI_FSM_STATE_OFF, MCI_FSM_STATE_REVERSE);
```
ShiCheng-Lu, line 32: fsm_transition takes fsm as first argument. (how did this compile?)

projects/mci/src/mci_fsm.c
```diff
@@ -0,0 +1,112 @@
...
+    } else {
+      LOG_DEBUG("didn't go to reverse\n");
+    }
+  } else if (notify_check_event(&notif, MCI_FSM_GOTO_CRUISE)) {
+    if (s_storage.velocity <= CRUISE_MAX_SPEED && s_storage.velocity >= CRUISE_MIN_SPEED) {
+      fsm_transition(fsm, MCI_FSM_STATE_CRUISE);
+    } else {
+      LOG_DEBUG("didn't go to drive\n");
```
ShiCheng-Lu, line 40: should be didn't go to cruise, or you can log something like "cruise requirement not met, didn't transition"

mitchellostler:
Overall Adhiraj, it looks good. You've got the state transitions correctly and fsms set up right. But there's a few things I want to discuss, and @ShiCheng-Lu It would be good to have your input as well. 

I think we really should just be checking the CAN messages directly (in the input functions) instead of using notifications. It's one less thing to go wrong, and then we don't need to worry about having another task notify the fsm. 
We have the `drive_output` message from centre console, and we can parse the value == whichever transition we want to process. For now we can just set this to an internal enum. 
Screen Shot 2022-11-01 at 9 24 53 PM
We have the can message definitions for MCI already from firmware 14, and they're set up for autogen so we can use the getters and setters for the drive output message and it should compile.

Also, in MCI for firmware_xiv, there is a guarded transition from off to on/reverse which checks that we are in the power state of MCI precharge charged. Since we don't have support guarded transitions anymore, this needs to be manually checked in the parse input function for the off state. I'm ok to leave it out for now but we should at least have a TODO

libraries/codegen/templates/_getters.h.jinja
```diff
@@ -12,3 +12,4 @@
     g_rx_struct.{{message}}_{{signal}}
     {%- endfor -%}
 {% endfor -%}
+{{'\n'}}
```
devAdhiraj, line 15: @JarvisWeng I had to add this line here to get rid of an error with auto gen.

projects/motor_controller/src/main.c
```diff
@@ -0,0 +1,40 @@
...
+static CanStorage s_can_storage = { 0 };
+const CanSettings can_settings = {
+  .device_id = 0x1,
+  .bitrate = CAN_HW_BITRATE_500KBPS,
+  .tx = { GPIO_PORT_A, 12 },
+  .rx = { GPIO_PORT_A, 11 },
+  .loopback = true,
+};
+
+TASK(master_task, TASK_STACK_512) {
```
JarvisWeng, line 20: There's an autogenerated template for this. Should follow that kind of platform. Look up the how to generate a new project

projects/motor_controller/src/mci_fsm.c
```diff
@@ -0,0 +1,114 @@
...
+
+static void prv_mci_fsm_reverse_input(Fsm *fsm, void *context) {
+  DriveOutputEvent drive_event = get_drive_output();
+  if (drive_event == MCI_FSM_GOTO_OFF) {
+    fsm_transition(fsm, MCI_FSM_STATE_OFF);
+  } else if (drive_event == MCI_FSM_GOTO_DRIVE) {
+    if (s_storage.velocity == 0) {
+      fsm_transition(fsm, MCI_FSM_STATE_DRIVE);
+    } else {
+      LOG_DEBUG("drive requirement not met, didn't transition\n");
```
JarvisWeng, line 59: Do you not need to send CAN messages?

projects/power_select/inc/power_select_power_supply_task.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
JarvisWeng, line 1: Why do you have `power_select` stuff in here? Did you `git merge`?

projects/motor_controller/src/main.c
mitchellostler: Will generating a new project overwrite this one?

mitchellostler: Will creating a new project overrwrite this one?

projects/motor_controller/src/mci_fsm.c
mitchellostler: I think this might be the next part of the project, but I would ask @ShiCheng-Lu . 
I'm thinking it would need to send an ack to centre and then store the state in some global storage for the motor controller task to access

projects/power_select/inc/power_select_power_supply_task.h
devAdhiraj: No, I tried to do git rebase main so I could get the new changes in this branch as they fixed some of the can autogen issues.

libraries/codegen/templates/_getters.h.jinja
```diff
@@ -11,4 +11,5 @@
 #define get_{{message}}_{{signal}}() \
     g_rx_struct.{{message}}_{{signal}}
     {%- endfor -%}
-{% endfor -%}
+{% endfor %}
```
devAdhiraj, line 14: Hey @JarvisWeng, I updated it to what you suggested and it seems to be working.

projects/motor_controller/src/main.c
devAdhiraj: Hey @JarvisWeng, I updated main.c to follow the auto generated template.

projects/motor_controller/src/mci_fsm.c
ShiCheng-Lu: thats another ticket

projects/motor_controller/src/main.c
JarvisWeng: Uh, i think it just appends it to the end of the file.
Oh, I meant the master task looks different

projects/power_select/inc/power_select_power_supply_task.h
JarvisWeng: Hm, I think you might have done it incorrectly then since you shouldn't be getting power select stuff then. Or you might have just shoved a the `power_select` commit in the middle of your branch.
If you have time, maybe you could move commits around so you don't need to have power select stuff

PR #105, opened 2022-10-29
Fwxv 134 create pedal main task

Scrubpai:


projects/pedal/src/main.c
```diff
@@ -0,0 +1,129 @@
...
+
+  pedal_calibrate();
+}
+
+void run_medium_cycle() {
+  run_can_tx_cycle();
+  wait_tasks(1);
+
+  // Sending messages
+  if (read_pedal_brake()) {
```
Bafran, line 79: We can consolidate this a bit, when brake is not pressed is the only time we have to send proper throttle data, so we can do something like
```
if (!read_pedal_brake()) { ... }
else { /* Send throttle data with brake as 0 */ }
```
also, throttle will be sent as per the reading from `read_throttle_data` so that should be changed from a bool to instead return an `uint16_t`.

projects/pedal/src/main.c
```diff
@@ -0,0 +1,129 @@
...
+
+#ifdef MS_PLATFORM_X86
+#define MASTER_MS_CYCLE_TIME 100
+#else
+#define MASTER_MS_CYCLE_TIME 1000
+#endif
+
+static CanStorage s_can_storage = { 0 };
+const CanSettings can_settings = {
+  .device_id = 0x1,
```
Bafran, line 25: This might end up changing again but leave the device id as 0x06 for now

projects/pedal/src/main.c
```diff
@@ -0,0 +1,129 @@
+#include <stdio.h>
+// #include "ads1015.h"
+#include "adc.h"
+#include "gpio.h"
+#include "gpio_it.h"
+#include "i2c.h"
+#include "interrupt.h"
+#include "soft_timer.h"
+// #include "event_queue.h"
```
Bafran, line 9: You can get rid of `#include "event_queue.h"`

Bafran (changes requested):
Overall looks good! For the whole main.c file, go through the unimplemented functions and just make sure there's a TODO comment explaining what it is as well. Also, `can.c`, `generator.py`, and `_rx_all.c.jinja` don't need to be included in this PR, so revert those changes and make sure the only three files changed are within the pedal project.

projects/pedal/src/main.c
mitchellostler: This is actually defined in `./libraries/ms-common/inc/can_board_ids.h` as SYSTEM_CAN_DEVICE_PEDAL

can/src/can.c
```diff
@@ -131,8 +131,6 @@ StatusCode can_transmit(const CanMessage *msg)
 {
   if (s_can_storage == NULL) {
     return status_code(STATUS_CODE_UNINITIALIZED);
-  } else if (msg->id.msg_id >= CAN_MSG_MAX_IDS) {
```
mitchellostler, line 134: Was there a reason for this being removed?

Bafran: Yeah, I was going to bring this up to Jarvis today, the linter sees this as a case that will never be hit because of the sizes of `msg->id.msg_id` and `CAN_MSG_MAX_IDS`. I assume the fix shouldn't be included in this PR though

JarvisWeng (approved):
Yeah I'm good with it. The `calib` can be another ticket I guess then

Bafran (approved):

PR #108, opened 2022-11-10
Add ads1015 files for x86

Bafran:
This is nearly the same as the files for x86 on FWXIV, just had to change the soft timer to work with the updated library.

### DIFF
--- a/projects/pedal/src/ads1015.c
+++ b/projects/pedal/src/ads1015.c
@@ -5,9 +5,10 @@
 // rotation of the channels is implemented using bitsets. The readings are just
 // a arbitrary value.
 #include "ads1015.h"
-#include <string.h>
-#include "FreeRTOS.h"
 
+#include <string.h>
+
+#include "FreeRTOS.h"
 #include "ads1015_def.h"
 #include "soft_timer.h"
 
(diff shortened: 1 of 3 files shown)


projects/pedal/src/ads1015.c
```diff
@@ -0,0 +1,126 @@
...
+#include <string.h>
+
+#include "FreeRTOS.h"
+#include "ads1015_def.h"
+#include "soft_timer.h"
+
+#define ADS1015_CHANNEL_UPDATE_PERIOD_US ADS1015_CONVERSION_TIME_US_1600_SPS
+#define ADS1015_CHANNEL_ARBITRARY_READING 0
+
+static SoftTimer s_timer;
```
JarvisWeng, line 18: Where does `s_timer` get initialized?

JarvisWeng (commented):
Other than the `s_timer` thing, I think it looks good. This will very much break with `x86` vs `arm`, but its ok for now

projects/pedal/src/ads1015.c
Bafran: Does it need to get initialized? There isn't an init function and `test_soft_timer.c` doesn't do any initialization either

JarvisWeng (approved):

PR #112, opened 2022-11-22
Nihal Sahu FW Training HW

nihal-sahu:
Here are my deliverables for FW 101, 102, and 103. I would appreciate any feedback :)

The FW103 tasks can be found in the freertos_task_implementation project directory.


nihal-sahu:
Just wanted to add that the build may be failing as the FW 102 tasks use a different branch. From what I recall from the repo it's missing some FreeRTOS stuff

projects/freertos_task_implementation/src/main.c
```diff
@@ -0,0 +1,66 @@
...
+  LOG_DEBUG("Task 1 initialized!\n");
+  StatusCode ret;
+  int i = 0;
+  while (true) {
+    ret = queue_send(&s_queue1, &s_list[i], 0);
+    delay_ms(100);
+
+    if (ret != STATUS_CODE_OK) LOG_DEBUG("write to queue failed\n");
+
+    i = (i < 5) ? (i + 1) : 0;
```
JarvisWeng, line 36: Oh instead of this you can use `i = i % 5`;

projects/freertos_task_implementation/tasks/tasks.c
```diff
@@ -0,0 +1,51 @@
...
+    prv_delay(1000);
+  }
+}
+
+int main(void) {
+  log_init();
+  tasks_init();
+
+  // Create tasks here
+  tasks_init_task(task1, TASK_PRIORITY(1), NULL);
```
JarvisWeng, line 43: Just as a question, why do these need to be different task priorities?

JarvisWeng (commented):
Yup looks good. Great work. I think you could probably take a ticket on standup this Saturday

nihal-sahu:
Awesome, thanks, should I close the PR then?

projects/freertos_task_implementation/tasks/tasks.c
nihal-sahu: I believe that the second part of this task was to make both tasks run while on different priorities.

PR #113, opened 2022-11-22
Fwxv 120 pwm x86 tests

harsh-patell:


libraries/ms-common/src/x86/pwm.c
```diff
@@ -5,18 +5,66 @@
...
+  } else if (pulse_width_ms > pwm[timer].period) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Pulse width must be leq period.");
+  }
+
+  pwm[timer].duty = (pulse_width_ms / pwm[timer].period) * 100;
+
+  return STATUS_CODE_OK;
 }
 
 StatusCode pwm_set_dc(PwmTimer timer, uint16_t dc) {
```
mitchellostler, line 47: You should be able to just store the duty cycle here, (`pwm[timer].duty = dc`) instead of converting to pulse and then setting the pulse (which

mitchellostler (commented):
Looks great, just the one fix to make things simpler and then we can merge. Good work, let me know if you have any questions

mitchellostler (commented):
Hey, sorry for how late this feedback is. 
The can_board_ids.h file is still in this PR, but it shouldn't be committed. Just do a 
`rm libraries/ms-common/inc/can_board_ids.h` from `fwxv/`

mitchellostler (approved):

PR #114, opened 2022-11-24
Fwxv 138 steering steering digital task

Citrusboa:


projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,97 @@
...
+                                                  DIGITAL_SIGNAL_CC_DECREASE_MASK);
+      }
+      break;
+
+    default:
+      return;
+  }
+}
+
+StatusCode steering_digital_input_init(void) {
```
Kyle02, line 82: Do you set any initial values for the different signals? They should probably be set to some default so that your functions are pulling a NULL input.

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,97 @@
...
+      if (state == GPIO_STATE_HIGH) {
+        set_digital_signal_regen_brake_toggle_command(1);
+      } else {
+        set_digital_signal_regen_brake_toggle_command(0);
+      }
+      break;
+
+    case STEERING_DIGITAL_INPUT_CC_TOGGLE:
+      if (state == GPIO_STATE_HIGH) {
+        set_digital_signal_cruise_control_command(CRUISE_CONTROL_COMMAND |
```
Kyle02, line 49: 000 | 0x03 -> 000 | 011 = 011. According to your docs, you want 100 as the result assuming CRUISE_CONTROL_COMMAND = 000.

Kyle02: same issue with the else statement of this case

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,97 @@
...
+                                                  DIGITAL_SIGNAL_CC_TOGGLE_MASK);
+      } else {
+        set_digital_signal_cruise_control_command(CRUISE_CONTROL_COMMAND ^
+                                                  DIGITAL_SIGNAL_CC_TOGGLE_MASK);
+      }
+      break;
+
+    case STEERING_DIGITAL_INPUT_CC_INCREASE_SPEED:
+      if (state == GPIO_STATE_HIGH) {
```
Kyle02, line 58: Probably should add an if statement checking to see that CC is on.

Kyle02 (changes requested):
Looks good, just some minor changes needed.

mitchellostler (commented):
1. Add gpio_it_register_interrupt for all necessary pins 
2. Read notification value in run_steering_digital_task
3. loop on  event_from_notification to parse each event and use in a case statement

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
...
+#define REGEN_BRAKE_COMMAND g_tx_struct.digital_signal_regen_brake_toggle_command
+#define CRUISE_CONTROL_COMMAND g_tx_struct.digital_signal_cruise_control_command
+
+static uint32_t notification = 0;
+Event steering_event;
+
+void run_steering_digital_task() {
+  notify_get(&notification);
+
+  while (event_from_notification(&notification, &steering_event)) {
```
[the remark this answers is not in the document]
Citrusboa: notify.h library comments updated. 
** any INVALID_EVENT returns STATUS_CODE_INVALID_ARGS in the handle_state_change()

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
...
+
+#define HORN_COMMAND g_tx_struct.digital_signal_horn_state
+#define REGEN_BRAKE_COMMAND g_tx_struct.digital_signal_regen_brake_toggle_command
+#define CRUISE_CONTROL_COMMAND g_tx_struct.digital_signal_cruise_control_command
+
+static uint32_t notification = 0;
+Event steering_event;
+
+void run_steering_digital_task() {
+  notify_get(&notification);
```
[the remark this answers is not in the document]
Citrusboa: got it

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
...
+
+  switch (digital_input) {
+    case STEERING_INPUT_HORN_EVENT:
+
+      set_digital_signal_horn_state(HORN_COMMAND ^ 1u);
+      break;
+
+    case STEERING_REGEN_BRAKE_EVENT:
+
+      set_digital_signal_regen_brake_toggle_command(REGEN_BRAKE_COMMAND ^ 1u);
```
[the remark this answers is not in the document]
Citrusboa: set all gpios to falling edge interrupt settings, besides horn which is rising_falling.

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
+#include "steering_digital_task.h"
+
+#define HORN_COMMAND g_tx_struct.digital_signal_horn_state
+#define REGEN_BRAKE_COMMAND g_tx_struct.digital_signal_regen_brake_toggle_command
+#define CRUISE_CONTROL_COMMAND g_tx_struct.digital_signal_cruise_control_command
+
+static uint32_t notification = 0;
+Event steering_event;
```
mitchellostler, line 8: should be declared static

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
...
+      set_digital_signal_horn_state(HORN_COMMAND ^ 1u);
+      break;
+
+    case STEERING_REGEN_BRAKE_EVENT:
+
+      set_digital_signal_regen_brake_toggle_command(REGEN_BRAKE_COMMAND ^ 1u);
+      break;
+
+    case STEERING_CC_TOGGLE_EVENT:
+
```
mitchellostler, line 33: Nit: lose the unecessary spacing, just a newline after each break will do

mitchellostler (changes requested):
Some minor functional changes, a couple nits.
The other thing, is can we change the names of steering_digital_task to steering_digital_input everywhere that it comes up? Since it's not a task, it's kind of a misnomer. 

There's also some changes that need to be made to how we set the events in the can message, but will need to confirm some things at standup today (will add comments when I do)

libraries/codegen/boards/steering.yaml
```diff
@@ -10,28 +10,16 @@
 
 ---
   Messages:
-    lights:
-      id: 24
+    digital_signal:
```
mitchellostler, line 13: We really only need to have one message for steering, so just call this one steering_info. 
I think it also makes sense to compress the digital inputs into one U8 (a signal 'digital_input'), with masks for each of the bits depending on what it means. ie, our U8 signal would look like the following where each item is a single bit

```
Messages:
  steering_info:
    id: xx
    target: xx
    signals:
      digital_input: 
        length: 8
```
 
 This digital_input byte will look like: 
 MSB   ->  LSB
 | (X) | (X) | (X) | (X) | (horn_state) |  (regen toggle state) | (cc_toggle_state) | (cc_inc_pressed) | (cc_decr_pressed) |

As this message is sent every cycle, we just need to update it whenever things change.

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
+#include "steering_digital_task.h"
+
+#define HORN_COMMAND g_tx_struct.digital_signal_horn_state
+#define REGEN_BRAKE_COMMAND g_tx_struct.digital_signal_regen_brake_toggle_command
+#define CRUISE_CONTROL_COMMAND g_tx_struct.digital_signal_cruise_control_command
+
+static uint32_t notification = 0;
+Event steering_event;
+
+void run_steering_digital_task() {
```
mitchellostler, line 10: We should zero the cc_decrease_pressed/cc_increase_pressed bits in the can message on every cycle, since we only want to transmit that that those events have happened once, as opposed to the toggle state messages where we should be broadcasting what the current state is.

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+StatusCode handle_state_change(Event digital_input) {
+  GpioState toggle_state;
+
+  switch (digital_input) {
+    case STEERING_INPUT_HORN_EVENT:
+
+      set_digital_signal_horn_state(HORN_COMMAND ^ 1u);
```
mitchellostler, line 24: Should become set_steering_info_digital_input(DIGITAL_INFO ^ HORN_STATE_MASK) as per comment above

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,74 @@
...
+      set_digital_signal_regen_brake_toggle_command(REGEN_BRAKE_COMMAND ^ 1u);
+      break;
+
+    case STEERING_CC_TOGGLE_EVENT:
+
+      set_digital_signal_cruise_control_command(CRUISE_CONTROL_COMMAND ^
+                                                DIGITAL_SIGNAL_CC_TOGGLE_MASK);
+      break;
+
+    case STEERING_INCREASE_SPEED_EVENT:
```
mitchellostler, line 38: This should just set the STEERING_CC_INCREASE_SPEED bit to 1 in the message, you don't need to read it back

projects/steering/src/steering_digital_task.c
```diff
@@ -0,0 +1,100 @@
...
+    case STEERING_REGEN_BRAKE_EVENT:
+
+      set_steering_info_digital_input(DIGITAL_INPUT ^ DIGITAL_SIGNAL_REGEN_BRAKE_MASK);
+      break;
+    case STEERING_CC_TOGGLE_EVENT:
+
+      set_steering_info_digital_input(DIGITAL_INPUT ^ DIGITAL_SIGNAL_CC_TOGGLE_MASK);
+      break;
+    case STEERING_CC_INCREASE_SPEED_EVENT:
+      toggle_state =
```
mitchellostler, line 52: Hey sorry, this still needs to be fixed. 
If the event occurs, the bit just needs to be set (just `set_steering_info_digital_input(DIGITAL_INPUT | DIGITAL_SIGNAL_CC_INCREASE_MASK)` You don't need to check the state, same goes for the decrease one below.

mitchellostler (commented):
Just the one change, otherwise looks good

mitchellostler (approved):

PR #115, opened 2022-11-27
Implemented brake_data and throttle_data

Scrubpai:


can/src/can.c
```diff
@@ -131,8 +131,6 @@ StatusCode can_transmit(const CanMessage *msg)
 {
   if (s_can_storage == NULL) {
     return status_code(STATUS_CODE_UNINITIALIZED);
-  } else if (msg->id.msg_id >= CAN_MSG_MAX_IDS) {
-    return status_msg(STATUS_CODE_INVALID_ARGS, "CAN: Invalid message ID");
```
Bafran, line 135: @JarvisWeng can this get merged with this PR? It was failing to build because the condition is always false, we had the same issue on the previous ticket but the fix was never merged

Bafran, line 135: @JarvisWeng can this get merged with this PR? It was failing to build because the condition is always false, we had the same issue on the previous ticket but the fix was never merged

projects/pedal/src/pedal_data.c
```diff
@@ -0,0 +1,43 @@
...
+#define THROTTLE_CHANNEL ADS1015_CHANNEL_0
+#define BRAKE_CHANNEL ADS1015_CHANNEL_2
+#define EE_PEDAL_VALUE_DENOMINATOR ((1 << 12))
+Ads1015Storage s_ads1015_storage = { 0 };
+
+// TODO(Scrubpai): Read any GPIOs and the ADC for throttle
+StatusCode read_throttle_data(int16_t *position) {
+  status_ok_or_return(ads1015_read_raw(&s_ads1015_storage, THROTTLE_CHANNEL, position));
+  int32_t position_upscaled = (int32_t)*position * EE_PEDAL_VALUE_DENOMINATOR;
+  int32_t range = 0;
```
Bafran, line 18: We should keep this as 1 for now since we're not editing it, otherwise the position will have issues with division by 0 on line 22. Same thing for the brake data function

projects/pedal/src/main.c
```diff
@@ -72,10 +65,15 @@ void run_medium_cycle() {
...
+  read_brake_data(&brake_position);
+  read_throttle_data(&throttle_position);
   // Sending messages
-  if (!read_pedal_brake()) {  // Brake is not pressed - Send proper throttle data
+  if (!brake_position) {  // Brake is not pressed - Send proper throttle data
     set_pedal_output_brake_output(0);
-    set_pedal_output_throttle_output(read_pedal_throttle());
+    set_pedal_output_throttle_output((uint32_t)throttle_position);
   } else {  // Brake is pressed - Send proper brake data with throttle as 0
     set_pedal_output_brake_output(1);
```
Bafran, line 78: Let's send the brake data we received when the brake is pressed, it likely won't be used but that can be decided on the motor controller's side.

Bafran (approved):

PR #118, opened 2022-12-03
transmit can msgs from mci to centre console

devAdhiraj:


libraries/codegen/boards/motor_controller.yaml
```diff
@@ -63,23 +63,42 @@
...
-        dsp_board_temp_r:
+        dsp_temp_r:
           length: 32
+
+    mc_status:
+      id: 40
+      target:
+        - centre_console
+      signals:
+        limit_bitset_left:
```
ShiCheng-Lu, line 92: probably want these to also be l/r instead of left/right for consistency, or change the others to left/right

projects/motor_controller/src/motor_controller_transmit_can_msgs_task.c
```diff
@@ -0,0 +1,63 @@
...
+  while (true) {
+    xSemaphoreTake(s_motor_controller_transmit_can_msgs_sem_handle, portMAX_DELAY);
+    counter++;
+    run_motor_controller_transmit_can_msgs_fast_cycle();
+    if ((counter % 10) == 0) run_motor_controller_transmit_can_msgs_medium_cycle();
+    if ((counter % 100) == 0) run_motor_controller_transmit_can_msgs_slow_cycle();
+  }
+}
+
+void run_motor_controller_transmit_can_msgs_fast_cycle() {}
```
ShiCheng-Lu, line 30: we can probably remove the cycles that are not used.

ShiCheng-Lu (commented):
Looks good to me.

ShiCheng-Lu (approved):

PR #120, opened 2022-12-30
Pedal calibrate

A3sanjay:


projects/pedal/src/flash.c
```diff
@@ -0,0 +1,101 @@
+#include "flash.h"
```
Bafran, line 1: This file is just the implementation for x86 - could you make two new directories under pedal, `pedal/src/x86` and `pedal/src/arm`, and then implement the arm version of flash.c as well (and move this file to x86 and the new one in arm). This should be pretty quick since it's also on FWIXV.

After that you should be able to run the calib test on both arm and x86, but you won't see any output for arm, just that the test passed.

Bafran (changes requested):
One file that needs to change but other than that just a few small things. Most of it looks good!

Bafran (approved):

PR #121, opened 2023-01-04
add mcp2515 to mci project

ShiCheng-Lu:


projects/motor_controller/src/mcp2515_hw.c
```diff
@@ -0,0 +1,395 @@
...
+  [CAN_HW_BITRATE_500KBPS] = MCP2515_CAN_BRP_500KBPS,
+};
+
+// SPI commands - See Table 12-1
+static void prv_reset() {
+  uint8_t payload[] = { MCP2515_CMD_RESET };
+  spi_exchange(s_storage->spi_port, payload, sizeof(payload), NULL, 0);
+
+  // this can cause problems maybe?
+  delay_ms(1);
```
mitchellostler, line 97: This shouldn't be necessary, since the spi_exchange call will block until the txn is complete

projects/motor_controller/src/mcp2515_hw.c
```diff
@@ -0,0 +1,395 @@
...
+    // RX overflow - clear error flags
+    uint8_t clear = 0;
+    prv_write(MCP2515_CTRL_REG_EFLG, &clear, 1);
+  }
+
+  s_storage->errors.eflg = err_flags;
+  prv_read(MCP2515_CTRL_REG_TEC, &s_storage->errors.tec, 1);
+  prv_read(MCP2515_CTRL_REG_REC, &s_storage->errors.rec, 1);
+
+  // keep error callback?
```
mitchellostler, line 185: I think we probably want to include this. At the very least, we should be notifying centre console/sending over telemetry so that these errors don't go unnoticed

projects/motor_controller/src/mcp2515_hw.c
```diff
@@ -0,0 +1,395 @@
...
+    prv_read(MCP2515_CTRL_REG_CANINTF, (uint8_t *)&regs, sizeof(regs));
+    // Mask out flags we don't care about
+    regs.canintf &= MCP2515_CANINT_EFLAG | MCP2515_CANINT_RX0IE | MCP2515_CANINT_RX1IE;
+
+    // Either RX or error
+    prv_handle_rx(regs.canintf);
+    prv_handle_error(regs.canintf, regs.eflg);
+    // critical_section_end(disabled);
+  }
+}
```
mitchellostler, line 211: It is true that notify_wait should never return false with block_indefinitely set, but I'm a bit uncomfortable leaving it like this since if this task exits it will crash the program. Can we do this in a while(true) with the notify_wait inside instead?

projects/motor_controller/src/mcp2515.c
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  return mcp2515_hw_transmit(msg->id.raw, msg->extended, msg->data, msg->dlc);
+}
+
+StatusCode mcp2515_init(Mcp2515Storage *storage, const Mcp2515Settings *settings) {
+  memset(storage, 0, sizeof(*storage));
+  s_storage = storage;
+
+  mcp2515_hw_init(&storage->rx_queue, settings);
```
mitchellostler, line 116: Thoughts on moving this into the initialization of one of the rx/tx tasks (with a lock until completion)? It needs the scheduler to be started for SPI, and unless we're planning to call this from the master task, it would be nice if we can just call mcp2515_init wherever, and it will handle setting itself up once the scheduler starts

mitchellostler (commented):
Overall looks good on the hw side. Based on the SPI design, we shouldn't run into any race conditions/timing issues, and I think that's the only place where there could be those kinds of issues

can/src/can.c
```diff
@@ -102,7 +102,7 @@ StatusCode can_init(CanStorage *storage, const CanSettings *settings)
   // Initialize hardware settings
   status_ok_or_return(can_hw_init(&s_can_storage->rx_queue, settings));
 
-  if (settings->mode == 0){
+  if (settings->mode == CAN_CONTINUOUS){
```
JarvisWeng, line 105: Thanks for adding this

projects/motor_controller/inc/mcp2515_defs.h
```diff
@@ -0,0 +1,145 @@
+#pragma once
```
JarvisWeng, line 1: I'll skip this file

projects/motor_controller/src/mcp2515.c
```diff
@@ -0,0 +1,139 @@
...
+
+  return STATUS_CODE_OK;
+}
+
+// should be auto gen
+void mcp2515_rx_all() {
+  CanMessage msg;
+
+  while (can_receive(&msg) == STATUS_CODE_OK) {
+    switch (msg.id.raw) {
```
JarvisWeng, line 44: You'll need a struct to populate the `msg`.
I think instead of autogen, you should just make one in here.

projects/motor_controller/src/mcp2515_hw.c
```diff
@@ -0,0 +1,395 @@
+
+#include "mcp2515_hw.h"
```
JarvisWeng, line 2: Seems very specific. I'll leave Mitchell to do this file.

projects/motor_controller/src/mcp2515.c
```diff
@@ -0,0 +1,139 @@
...
+    // LOG_DEBUG("Data: %lx\n", msg->data);
+    // LOG_DEBUG("DLC: %ld\n", msg->dlc);
+    // LOG_DEBUG("ret: %d\n", ret);
+  }
+
+  return ret;
+}
+
+// should be auto gen
+void mcp2515_tx_all() {
```
JarvisWeng, line 77: Same here. Just make a `tx_struct`. And then just send out the messages manually.
Though if you just want autogen first, then just copy paste I can show you how to do that

projects/motor_controller/src/mcp2515_hw.c
```diff
@@ -0,0 +1,395 @@
...
+      rx_msg.id.raw = id.raw;
+
+      uint8_t data_payload[] = { MCP2515_CMD_READ_RX | rx_buf->data };
+      uint64_t read_data = 0;
+      spi_exchange(s_storage->spi_port, data_payload, sizeof(data_payload), (uint8_t *)&rx_msg.data,
+                   sizeof(rx_msg.data));
+      // Clear the interrupt flag so a new message can be loaded
+      prv_bit_modify(MCP2515_CTRL_REG_CANINTF, rx_buf->int_flag, 0x0);
+
+      can_queue_push(&s_storage->rx_queue, &rx_msg);
```
mitchellostler, line 163: @JarvisWeng Does it make sense to just push these on to the global can_queue and process them like normal can messages? or is that going to be too much of a pain?

ShiCheng-Lu: it should be easier to keep the queues separate since the can lines are separate, so there won't be id conflicts.

projects/motor_controller/src/mcp2515.c
ShiCheng-Lu: there are code here that require scheduler to not be started, I've moved the portion that need the scheduler into the Interrupt task, which has a higher priority, so it should be ran before RX/TX. (unless we really want a lock to protect it this should be fine)

mitchellostler (approved):
LGTM, good work. What's your plan for the mcp2515_tx/rx_all functions? are they going to be user defined per board or just implemented later?

projects/motor_controller/src/mcp2515.c
ShiCheng-Lu: changed rx_all and tx_all to be function pointers, since the can definitions are not hw related, I'd like it to be done in another pr.

projects/motor_controller/src/mcp2515.c
```diff
@@ -0,0 +1,125 @@
...
+}
+
+static void no_op() {}
+
+StatusCode mcp2515_init(Mcp2515Storage *storage, const Mcp2515Settings *settings) {
+  memset(storage, 0, sizeof(*storage));
+  s_storage = storage;
+
+  if (mcp2515_rx_all == NULL) {
+    mcp2515_rx_all = no_op;
```
JarvisWeng, line 95: Lol I guess I'm fine with this for now.

JarvisWeng (approved):

PR #124, opened 2023-01-25
Add centre console project 😉

mitchellostler:


projects/centre_console/config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "libs": [
+        "FreeRTOS",
+        "ms-common"
+    ]
```
JarvisWeng, line 5: You could add `can: True` but then you would need to make a yaml too so whatever

JarvisWeng (approved):

PR #125, opened 2023-01-28
Fwxv 130 power select tests & Refactoring

03predo:
Refactored some of the power select code to all follow the same format. Implemented a simple test in task for the power select FSMs.


03predo:
For some reason when I split up the tests the third always fails. I have switched the order and its always the third one failing.

JarvisWeng:
> For some reason when I split up the tests the third always fails. I have switched the order and its always the third one failing.

Hm not too sure. @mitchellostler any ideas?

projects/power_select/test/test_power_select.c
```diff
@@ -0,0 +1,93 @@
...
+#include "task_test_helpers.h"
+#include "unity.h"
+
+void setup_test(void) {
+  log_init();
+}
+
+void teardown_test(void) {}
+
+TEST_IN_TASK
```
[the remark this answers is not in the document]
ShiCheng-Lu: runs the test in task context, runs after scheduler has started.

JarvisWeng (commented):
Hm, seems like there's a lot of changes already on `main`. Could you possibly `rebase` to make it clear which changes are yours?

03predo:
> Hm, seems like there's a lot of changes already on `main`. Could you possibly `rebase` to make it clear which changes are yours?

projects/power_select/test/test_power_select_aux_bat.c
```diff
@@ -0,0 +1,47 @@
...
+  wait_tasks(1);
+  TEST_ASSERT_EQUAL(aux_bat_fsm->curr_state->id, POWER_SELECT_INACTIVE);
+  TEST_ASSERT_EQUAL(g_tx_struct.power_select_status_status, 0);
+
+  fsm_run_cycle(aux_bat);
+  wait_tasks(1);
+  TEST_ASSERT_EQUAL(aux_bat_fsm->curr_state->id, POWER_SELECT_INACTIVE);
+  TEST_ASSERT_EQUAL(g_tx_struct.power_select_status_status, 0);
+
+  vTaskDelay(pdMS_TO_TICKS(1000));
```
mitchellostler, line 46: You shouldn't need this here, have you tried without it? (same goes for other delays)
Basically, we know that the cycle has complete since we wait_tasks, so we can just exit the function after our last test_assert

mitchellostler (approved):
Tests look good in terms of format, but I would make sure you get @JarvisWeng's approval for all the power select changes.

libraries/ms-common/src/log.c
```diff
@@ -16,7 +16,9 @@ static uint8_t s_log_array[QUEUE_LENGTH * ITEM_SIZE];
 QueueHandle_t g_log_queue = NULL;
 
 void log_init(void) {
```
mitchellostler, line 18: Sorry, we just pushed some changes to our log library, so if you just get rid of these changes you shouldn't have any more conflicts

projects/power_select/test/test_power_select_aux_bat.c
03predo: Oh yeah the main reason for those was just to be able to see the log output but I can remove them

JarvisWeng (approved):
Yup looks good to me

PR #126, opened 2023-01-28
Add HW definitions, make LEDs work on main

mitchellostler:
Added the stuff I used to fix LEDs, so know scons flash --project=leds will work on master


libraries/ms-common/src/arm/gpio.c
```diff
@@ -28,6 +28,9 @@ static uint32_t s_gpio_rcc_apb_timer_map[NUM_GPIO_PORTS] = {
 };
 
 StatusCode gpio_init(void) {
+  // Remap pins to disable jtag and enable LEDs
```
JarvisWeng, line 31: Maybe just add a comment and link for why this is necessary

JarvisWeng (approved):

PR #127, opened 2023-01-28
created smoke directory with adc project

harsh-patell:


smoke/adc/src/main.c
```diff
@@ -0,0 +1,52 @@
...
+      uint16_t data = 0;
+      for (uint8_t i = 0; i < SIZEOF_ARRAY(adc_addy); i++) {
+        adc_read_converted(adc_addy[i], &data);
+        LOG_DEBUG("%d%d: %d\n", adc_addy[i].port, adc_addy[i].pin, data);
+      }
+      delay_ms(1000);
+   }
+}
+
+int main() {
```
mitchellostler, line 40: You need to add an adc_init(ADC_MODE_SINGLE) after the other initializations

mitchellostler (commented):
Just the one thing, otherwise looks good!

smoke/adc/src/main.c
harsh-patell: Added

harsh-patell: I'm not sure why CI fails on format and lint? There was no issue when I tested it

mitchellostler (approved):

PR #130, opened 2023-01-30
Fwxv 999 kevinsong 103 hw

kevinksong:
Last one was a mistake sorry, here is my fw 103 work


projects/queues/src/queues1.c
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  char outstr[ITEM_SZ];
+  StatusCode ret;
+  while (true) {
+    // Your code goes here
+    ret = queue_receive(&s_queue1, &outstr, 100);
```
Bafran, line 48: for the `queue_receive` function, the second parameter is a pointer to a buffer that the queue reads into, so you just need to pass in `outstr`, you don't need to pass in the &address of the array.

Bafran:
There's only one major issue I saw, you can make that change and try run it on your computer. Otherwise looks good and we can close this PR!

PR #136, opened 2023-02-03
FW103_Task3_Queues

wavy152:


queues/src/main.c
```diff
@@ -0,0 +1,74 @@
...
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  const char outstr[ITEM_SZ];
+  StatusCode ret;
+  while (true) {
+    // Your code goes here
+    for (size_t i = 0; i < QUEUE_LEN; i++) {
+      ret = queue_receive(&s_queue1, &outstr[i], 0);
```
Kyle02, line 50: delay ms needs to be 100 otherwise you timeout (can be any amount of time) 100 is just a random num

queues/src/main.c
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  const char outstr[ITEM_SZ];
+  StatusCode ret;
+  while (true) {
+    // Your code goes here
+    for (size_t i = 0; i < QUEUE_LEN; i++) {
```
Kyle02, line 49: You almost have it, just need to fix your task2. If you try running it you will get a weird output. You want something like the screenshot I just sent you. Try changing your task2 to try to get your output to match the image. Hint: Think about what you are sending through the queue (and therefore receiving). 
Screen Shot 2023-02-04 at 10 15 47 PM

Kyle02 (changes requested):

mitchellostler (commented):
Looks Good! If you come to the next standup, you can take on a task

PR #137, opened 2023-02-04
Power FSM

Bafran:


projects/centre_console/inc/power_fsm_sequence.h
```diff
@@ -0,0 +1,20 @@
+#include "fsm.h"
+
+void prv_power_fsm_confirm_aux_status_input(Fsm *fsm, void *context);
```
mitchellostler, line 3: Can we group functions by pair (input/output together), and then add a description for each state

projects/centre_console/src/power_fsm.c
```diff
@@ -0,0 +1,136 @@
...
+  // -> OFF Sequence
+  STATE(POWER_FSM_DISCHARGE_PRECHARGE, prv_power_fsm_discharge_precharge_input, prv_power_fsm_discharge_precharge_output),
+  STATE(POWER_FSM_TURN_OFF_EVERYTHING, prv_power_fsm_turn_off_everything_input, prv_power_fsm_turn_off_everything_output),
+  STATE(POWER_FSM_OPEN_RELAYS, prv_power_fsm_open_relays_input, prv_power_fsm_open_relays_output),
+};
+
+// Declares transition for state machine
+static FsmTransition s_power_fsm_transitions[NUM_CENTRE_CONSOLE_POWER_TRANSITIONS] = {
+  // Transitions for OFF state
+  TRANSITION(POWER_FSM_STATE_OFF, POWER_FSM_STATE_MAIN),
```
mitchellostler, line 71: So these transitions shouldn't exist since we never transition from OFF->MAIN, etc directly. We should only have our sequences

JarvisWeng (approved):

JarvisWeng:
This is just for the structure right?

mitchellostler (commented):
Can you create a diagram of this FSM if it's not too much of a pain in the ass? 
Would be really useful to have, and catch any issues with structure. I typically use lucidchart, but you can use any tool you want, and just pop it on the confluence page. (If you can save it too that's great because we may have to edit it later. I'm gonna see if we can get a plugin for confluence to make this easier)

projects/centre_console/inc/power_fsm_sequence.h
```diff
@@ -0,0 +1,46 @@
+#ifndef PROJECTS_CENTRE_CONSOLE_INC_POWER_FSM_SEQUENCE_H_
```
mitchellostler, line 1: #pragma once

mitchellostler:
> Can you create a diagram of this FSM if it's not too much of a pain in the ass? Would be really useful to have, and catch any issues with structure. I typically use lucidchart, but you can use any tool you want, and just pop it on the confluence page. (If you can save it too that's great because we may have to edit it later. I'm gonna see if we can get a plugin for confluence to make this easier)

Ignore this until we have a better solution since it's likely to change

PR #138, opened 2023-02-04
Implement PCA9555 Driver

03predo:
The PCA9555 has the same registers addresses and uses the i2c commands as PCA9539R. The only difference in implementation is how interrupts are handled with FreeRTOS now.


mitchellostler (approved):

mitchellostler:
Great work

PR #139, opened 2023-02-04
Fwxv 152 write max11600 driver

Scrubpai:


projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+// External Clock: 1
+// Unipolar: 0
+// Resetting Configuration: 0
+// Don't care: 0
+static const uint8_t max11600_setup_byte = 0b10001000;
+
+// Scanning Pattern: 00
+// Channel Select: AIN2
+// Single-ended: 1
+static const uint8_t max11600_config_byte = 0b00000101;
```
Bafran, line 19: Just a minor thing, move these to the config and setup variables to the .c file since these don't need to be used outside of that

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+
+// Read Mask
+static const uint8_t max11600_read_address = 0b11001001;
+// Write Mask
+static const uint8_t max11600_write_address = 0b11001000;
```
Bafran, line 24: And I think read and write addresses can be changed to a #define, also in the .c file, since we don't need to pass the address, just the value.

Bafran (approved):
Looks good, just minor stuff but it seems like the functionality should work, pending a hardware test. I think @mitchellostler wanted to take a look as well

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "gpio.h"
+#include "i2c.h"
+
+// Reference Voltage: 000
+// External Clock: 1
+// Unipolar: 0
+// Resetting Configuration: 0
+// Don't care: 0
+static const uint8_t max11600_setup_byte = 0b10001000;
```
mitchellostler, line 14: These should be #defines, and in hexidecimal form (ie for this one: `#define MAX11600_SETUP_BYTE 0x88`) Also, as adel mentioned move them to the .c file

The comments are great though, keep those. Although it doesn't look like they map directly to the value, am I reading that wrong?

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+// Single-ended: 1
+static const uint8_t max11600_config_byte = 0b00000101;
+
+// Read Mask
+static const uint8_t max11600_read_address = 0b11001001;
+// Write Mask
+static const uint8_t max11600_write_address = 0b11001000;
+
+#define NUM_MAX11600_CHANNELS 3
+#define REFERENCE_VOLTAGE 5
```
mitchellostler, line 27: Add units to this (_MV or _V depending on if it's millivolts or volts)

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+
+typedef struct {
+  I2CPort i2c_port;  // I2C_PORT_1 or I2C_PORT_2
+  I2CAddress i2c_read_address;
+  I2CAddress i2c_write_address;
+  int16_t channel_readings[NUM_MAX11600_CHANNELS];
+} Max11600Storage;
+
+// i2c_write(I2C_PORT_1, MAX11600_WRITE_ADDRESS, MAX11600_SETUP_BYTE, 1);
```
mitchellostler, line 36: Is there a reason this is here?

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+StatusCode max11600_init(Max11600Storage *Storage, I2CPort i2c_port, I2CAddress i2c_read_address,
```
mitchellostler, line 37: Can you add comments for each of the functions stating what they do, and if there's any specific configuration that is needed to call them, or any errors that they may return. Always good practice to document your code as best as possible

mitchellostler, line 37: Nit: Storage should be fully lowercase everywhere since it's a variable name

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
+#ifndef PROJECTS_PEDAL_INC_MAX11600_H_
```
mitchellostler, line 1: Use #pragma once and not ifndef

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+static const uint8_t max11600_write_address = 0b11001000;
+
+#define NUM_MAX11600_CHANNELS 3
+#define REFERENCE_VOLTAGE 5
+
+typedef struct {
+  I2CPort i2c_port;  // I2C_PORT_1 or I2C_PORT_2
+  I2CAddress i2c_read_address;
+  I2CAddress i2c_write_address;
+  int16_t channel_readings[NUM_MAX11600_CHANNELS];
```
mitchellostler, line 33: Is it specified that we get signed values from the sensor? if not we need to make sure that this is a uint.

How are the readings specified? Are they 16 bits? are they signed or unsigned?

projects/pedal/inc/max11600.h
```diff
@@ -0,0 +1,44 @@
...
+} Max11600Storage;
+
+// i2c_write(I2C_PORT_1, MAX11600_WRITE_ADDRESS, MAX11600_SETUP_BYTE, 1);
+StatusCode max11600_init(Max11600Storage *Storage, I2CPort i2c_port, I2CAddress i2c_read_address,
+                         I2CAddress i2c_write_address);
+
+StatusCode max11600_read_raw(Max11600Storage *Storage);
```
mitchellostler, line 40: Pretty important to say that this is updating the same storage readings with either the raw or converted value, since it's not clear from the interface

projects/pedal/src/max11600.c
```diff
@@ -0,0 +1,42 @@
...
+  Storage->channel_readings[0] = rx_data[0];
+  Storage->channel_readings[1] = rx_data[1];
+  Storage->channel_readings[2] = rx_data[2];
+  return status;
+}
+
+StatusCode max11600_read_converted(Max11600Storage *Storage) {
+  StatusCode status;
+  status = max11600_read_raw(Storage);
+  Storage->channel_readings[0] *= REFERENCE_VOLTAGE / 256 * 1000;
```
mitchellostler, line 38: This is going to be a problem, since this value will just become zero as the value is cast implicitly to an integer in 5 / 256 (no decimal points are stored). If you move the *1000 to the front that should resolve it. You should definitely indicate that the output is in mv in the function description.

projects/pedal/inc/max11600.h
mitchellostler: Also, where is this value from?

mitchellostler (changes requested):
In general, it looks really good. 
A few comments that should be addressed, then we can move it in

projects/pedal/inc/max11600.h
Bafran: You're right this should change to a uint8_t, the MAX is 8-bit and since we're in unipolar our readings are unsigned

projects/pedal/inc/max11600.h
Bafran: The I2C functions take a pointer to data that's why I thought it'd be best to leave them as variables although not sure what the cleanest way to do that is.

Also just for my own knowledge I thought defining the bytes in binary would be more clear given that each bit serves a specific purpose, is there a reason to keep everything hex?

Bafran (approved):

mitchellostler (approved):

PR #140, opened 2023-02-04
FWXV-181 Drive fsm created

Kyle02:


Kyle02:
Hey Mitchell, do you know if drive fsm is supposed to ever set an error state? I don't believe it is suppose to. I assume it just gets error from power fsm and should go to neutral when we error.

mitchellostler (commented):
Take a look at these comments if you get a chance, but I think we should meet to go through this together and discuss, hopefully tomorrow after standup

mitchellostler (commented):
- Add drive error state to shared memory, set if check_precharge fails
- remove transmit state, move precharge to one file: Neutral -> drive and neutral -> Reverse should both set state they're trying to transition to (next_state), and then go to precharge
- Drive state (in CAN message) should be set in the OUTPUT function of the state (D, N, R, only)

projects/centre_console/inc/fsm_shared_mem.h
```diff
@@ -5,6 +5,17 @@
...
+
+void fsm_shared_mem_init(FSMStorage *storage);
+void fsm_shared_mem_set_power_state(FSMStorage *storage, StateId state);
+StateId fsm_shared_mem_get_power_state(FSMStorage *storage);
+void fsm_shared_mem_set_power_error_code(FSMStorage *storage, StatusCode code);
+StatusCode fsm_shared_mem_get_power_error_code(FSMStorage *storage);
+void fsm_shared_mem_set_drive_error_code(FSMStorage *storage, StatusCode code);
+StatusCode fsm_shared_mem_get_drive_error_code(FSMStorage *storage);
+
+extern FSMStorage cc_storage;
```
mitchellostler, line 21: Just declare this in the .c file, since we'll only ever need one. You can remove the pointer from the methods too, and then just do all the operations on the single storage in the .c file

projects/centre_console/src/main.c
```diff
@@ -1,24 +1,61 @@
...
+static CanStorage s_can_storage = { 0 };
+const CanSettings can_settings = {
+  .device_id = 0x1,
+  .bitrate = CAN_HW_BITRATE_500KBPS,
+  .tx = { GPIO_PORT_A, 12 },
+  .rx = { GPIO_PORT_A, 11 },
+  .loopback = true,
+};
+
+// FSMStorage cc_storage = {0}; not sure why this is breaking, might be a linking issue
```
mitchellostler, line 29: remove this

projects/centre_console/src/main.c
```diff
@@ -1,24 +1,61 @@
...
 void run_fast_cycle() {}
 
-void run_medium_cycle() {}
+void run_medium_cycle() {
+  run_can_rx_cycle();
+  fsm_run_cycle(drive);
+  fsm_run_cycle(centre_console_power_fsm);
+  wait_tasks(2);
+  run_can_tx_cycle();
+  // delay_ms(1000);
```
mitchellostler, line 39: remove

projects/centre_console/src/main.c
```diff
@@ -1,24 +1,61 @@
...
 
 int main() {
-  tasks_init();
   log_init();
-  LOG_DEBUG("Welcome to TEST!");
+  tasks_init();
+  can_init(&s_can_storage, &can_settings);
+
+  LOG_DEBUG("Welcome to TEST! \n");
+  // fsm_shared_mem_init(&cc_storage);
```
mitchellostler, line 50: you can call fsm_shared_mem_init here once you make the other change

projects/centre_console/src/fsm_shared_mem.c
```diff
@@ -28,17 +29,32 @@ StateId fsm_shared_mem_get_power_state(FSMStorage *storage) {
...
   mutex_lock(&storage->mutex, BLOCK_INDEFINITELY);
-  StatusCode error = storage->error_code;
+  StatusCode error = storage->power_error_code;
+  mutex_unlock(&storage->mutex);
+  return error;
+}
+
+// set drive error_code
+void fsm_shared_mem_set_drive_error_code(FSMStorage *storage, StatusCode code) {
+  mutex_lock(&storage->mutex, BLOCK_INDEFINITELY);
```
mitchellostler, line 49: Definitely don't want to block indefinitely here, set to like 100 ms or something

mitchellostler (commented):
Hey, just a couple changes to fsm_shared memory (since I realized what the issues you were having were). 
You don't need to declare extern, just have a static instance in the fsm_shared_mem c file. Then you can call all the methods from wherever and just do the operations on that static member in the c file .

Otherwise LGTM!

mitchellostler (approved):

PR #142, opened 2023-02-07
Motor controller can

ShiCheng-Lu:


projects/motor_controller/src/motor_can.c
```diff
@@ -0,0 +1,109 @@
...
+  RESERVED,
+  HEAT_SINK_MOTOR_TEMP,
+  DSP_BOARD_TEMP,
+  RESERVED_2,
+  ODOMETER_BUS_AMPHOUR,
+  SLIP_SPEED = 0x17,
+} MotorControllerMessageIds;
+
+static void motor_controller_tx_all() {
+  // s_send_message will be false if no can message from center_console and peddle were received
```
mitchellostler, line 41: Is this what we want to do, since we need to communicate with the motor controllers every 250ms?

I guess if we have stale values, it's an issue, because that means we aren't sending pedal readings fast enough. I think as long as the transmission frequency from pedal/Centre console  is >= frequency we update the motor controllers, we theoretically shouldn't have stale values. 

In practice though, if an issue occurs and we have stale values, I would think that we would want to still transmit a message to the motors. If we are sending a cruise control value, it's fine to send the stale value again, but if we are sending torque we should probably send zero so it doesn't loop on the same values and continue to magnify itself. It's not ideal to send zero, but since this shouldn't happen, and the alternative (if we don't send a message) would be the motors shutting down, I think it should be ok.

projects/motor_controller/src/motor_can.c
```diff
@@ -0,0 +1,109 @@
...
+    switch (msg.id.raw) {
+      case MOTOR_CONTROLLER_BASE_L + STATUS:
+        set_motor_status_motor_status_l(msg.data_u32[1]);
+        break;
+      case MOTOR_CONTROLLER_BASE_R + STATUS:
+        set_motor_status_motor_status_r(msg.data_u32[1]);
+        break;
+
+      case MOTOR_CONTROLLER_BASE_L + BUS_MEASUREMENT:
+        set_motor_controller_vc_mc_current_l(uint32_as_float(msg.data_u32[0]) * CURRENT_SCALE);
```
mitchellostler, line 79: Are you going to change this so that we are sending all the data as uints?

mitchellostler (commented):
Overall looks good, I just have the two questions

projects/motor_controller/src/motor_can.c
ShiCheng-Lu: this will be updated with can watch dog once that gets done.

mitchellostler (approved):

PR #143, opened 2023-02-07
update ubuntu

ShiCheng-Lu:
github is deprecating ubuntu-18.04, which means older workflow CI checks cannot complete. update to ubunto 20.04 (which is also box ubuntu version, don't know why the CI wasn't updated) and fix issues with this update

### DIFF
--- a/libraries/test_gen/src/test_generator.py
+++ b/libraries/test_gen/src/test_generator.py
@@ -29,7 +29,7 @@ def get_function_line_spacing(function_words):
             spacing -= 1
 
     if bracket_location == -1:
-        raise Exception("No bracket in provided line!")
+        raise ValueError("No bracket in provided line!")
 
     # Count the number of letters leading up to the bracket
     for i in range(len(function_words[bracket_location])):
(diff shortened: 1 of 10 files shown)


libraries/test_gen/src/test_generator.py
```diff
@@ -29,7 +29,7 @@ def get_function_line_spacing(function_words):
             spacing -= 1
 
     if bracket_location == -1:
-        raise Exception("No bracket in provided line!")  # pylint: disable=W0719
+        raise Exception("No bracket in provided line!")
```
mitchellostler, line 32: We were running into a linting error on this line, that's why I got Ivy to add the disable comment. Did we update the version of pylint, or is it mismatched in the CI?

mitchellostler (approved):

libraries/test_gen/src/test_generator.py
ShiCheng-Lu: I'm assuming you guys were running into linting during CI but not commit, the CI pipeline was 18.04 so it had an older version of pylint. In this version it was complaining about that pylint comment being there instead.

ShiCheng-Lu:
also, I don't have permission to merge, so...

PR #148, opened 2023-02-11
can allow mismatched data length v2

ShiCheng-Lu:


ShiCheng-Lu:
I noticed that people were accessing g_rx_struct for data, these do not have getters. If we allow them to access the data, why even provide getters and setters. I don't think the getters and setters should be generated if we are going to be allow all read/write access anyway.

libraries/codegen/templates/_rx_all.c.jinja
```diff
@@ -12,18 +12,17 @@ void can_rx_all() {
...
             case SYSTEM_CAN_MESSAGE_{{message_data["src"] | upper}}_{{message | upper}}:
-
-    {%- set signals = message_data["signals"].keys() -%}
-    {%- set function_parameters -%}
-can_unpack_{{message}}(&msg {%- for signal in signals %}, &g_rx_struct.{{message}}_{{signal}} {%- endfor -%});
-    {%- endset %}
-                {{function_parameters | format_function_length(16)}}
+            {%- set offset = namespace(value=0) %}
+            {%- for signal, signal_data in message_data['signals'].items() %}
+                g_rx_struct.{{message}}_{{signal}} = (msg.data >> {{offset.value}});
```
JarvisWeng, line 19: What if the data is in the middle of the message? Also do I need to review `v1` as well?

libraries/codegen/templates/_tx_all.c.jinja
```diff
@@ -1,18 +1,26 @@
...
-{%- for message, message_data in messages.items() -%}
-    {%- set signals = message_data["signals"].keys() -%}
-    {%- set function_parameters -%}
-can_pack_{{message}}(&msg {%- for signal in signals %}, g_tx_struct.{{message}}_{{signal}} {%- endfor -%});
-    {%- endset %}
-
-    {{function_parameters | format_function_length(4)}}
+{%- for message, message_data in messages.items() %}
+    {%- set signals = message_data["signals"].keys() %}
+    can_pack_impl_u64(
```
JarvisWeng, line 14: wait you can't pack it in 64. Not every message will be 64 bits

libraries/codegen/templates/_tx_all.c.jinja
```diff
@@ -1,18 +1,26 @@
...
+        &msg, SYSTEM_CAN_MESSAGE_{{data["Board"] | upper}}_{{message | upper}}, 
+        {{- message_data['signals'].values() | sum(attribute='length') }},
+    {%- set offset = namespace(value=0) %}
+    {%- for signal, signal_data in message_data['signals'].items() %}
+        (uint64_t) g_tx_struct.{{message}}_{{signal}} << {{offset.value}} &
```
JarvisWeng, line 19: What's the `&` for? Again what about signals in the middle

ShiCheng-Lu: bit wise and, it shifts the message and stacks it on top
example in hex: uint8_t: aa, uint16_t: bbbb, uint32_t cccccccc
aa << 0 --> aa
bbbb << 8 --> bbbb00
cccccccc << 24 --> cccccccc000000

result: ccccccccbbbbaa
and with little endian, this is the first 7 bytes of the uint64_t data

libraries/codegen/templates/_tx_all.c.jinja
ShiCheng-Lu: when packed, it should take up the first however bit of the data that it needs

libraries/codegen/templates/_rx_all.c.jinja
ShiCheng-Lu: since g_rx_struct.message_signal is the length of the data, it should chop off on the most significant end. The right shift should end the data on the least significant side

JarvisWeng: Hm, but is this meant to just do `bytes` or any lengths?

libraries/codegen/templates/_getters.h.jinja
```diff
@@ -1,15 +1,13 @@
...
+{% for message, message_data in messages.items() %}
     {%- for signal, signal_data in message_data['signals'].items() %}
-
 #define get_{{message}}_{{signal}}() \
     g_rx_struct.{{message}}_{{signal}}
-    {%- endfor -%}
-{% endfor %}
-
+    {% endfor %}
+{%- endfor %}
```
JarvisWeng, line 13: If you're deleting `_pack_msg.h.jinja` and `_unpack_msg.h.jinja`, you don't need `can/src/can_pack_impl.c` or `can/src/can_unpack_impl.c` either right?

libraries/codegen/templates/_tx_all.c.jinja
JarvisWeng: hm, so you're just shoving concatenating it all into a `uint64_t` to pack

libraries/codegen/templates/_getters.h.jinja
JarvisWeng: You just need `can_pack_64u` but I guess you can just put that somewhere else

libraries/codegen/templates/_rx_all.c.jinja
JarvisWeng: Also, I feel that it would be better to use like the `data_u8` or `data_u16` instead of just shifting if it's a `byte`.
Would need more logic though

libraries/codegen/templates/_getters.h.jinja
ShiCheng-Lu: that is true.

libraries/codegen/templates/_rx_all.c.jinja
ShiCheng-Lu: you can't access bytes 8-23 as a uint16 with data_u16. the offset offsets the 64 bit data so that the start is aligned, then assigning onto the g_rx_struct will limit the end

libraries/codegen/templates/_tx_all.c.jinja
ShiCheng-Lu: yes

libraries/codegen/templates/_tx_all.c.jinja
JarvisWeng: Instead of & 0 don't you want to AND FFFF

ShiCheng-Lu: right

ShiCheng-Lu: These should all be ORs good catch, or else our messages would've been always empty

libraries/codegen/templates/_rx_all.c.jinja
ShiCheng-Lu: too complicated, and I don't think it's worth the effort, bit shift should be close to the same performance as accessing using the union type

JarvisWeng: Yeah it is a bit complicated. It just says a little computation time if you don't need to `bit shift` it but I'm ok with it

libraries/codegen/templates/_tx_all.c.jinja
```diff
@@ -1,18 +1,33 @@
...
 #include "can_codegen.h"
 
-void can_tx_all() {
-    CanMessage msg = { 0 };
-{%- for message, message_data in messages.items() -%}
-    {%- set signals = message_data["signals"].keys() -%}
-    {%- set function_parameters -%}
-can_pack_{{message}}(&msg {%- for signal in signals %}, g_tx_struct.{{message}}_{{signal}} {%- endfor -%});
-    {%- endset %}
+static CanMessage s_msg = {
```
JarvisWeng, line 9: I suppose this is fine. Though I just put it in the function so we could use the `stack`. Since we don't really use like any of our stack space. This is fine though

JarvisWeng (commented):
I'll have to pull the branch and generate the files just to see. But looks good so far. If you had no more changes, let me know when I could do that

JarvisWeng (approved):
Files look fine

PR #149, opened 2023-02-13
FWXV-166: UV cutoff task

devAdhiraj:


libraries/codegen/boards/uv_cutoff.yaml
```diff
@@ -0,0 +1,19 @@
...
+#
+# If you are making changes to this file please update the corresponding entry
+# on the wiki. If you need to add a new message use a reasonable
+# reserved ID. The higher ID the lower the priority. Generally
+# - 0-13: Critical messages (have ACK)
+# - 14-30: Actionable messages (trigger a change in another system)
+# - 30-63: Data messages (usually not actionable by an onboard device)
+
+---
+  Messages:
```
JarvisWeng, line 12: Wow, you got the CAN stuff. Nice!

JarvisWeng (approved):
This looks mostly good to me. Just fix up the build failure.
@mitchellostler what do you think?

libraries/codegen/boards/uv_cutoff.yaml
mitchellostler: > This looks mostly good to me. @mitchellostler what do you think?
What's the scope of this ticket? Do we do any interfacing with the IC? I'm confused as to what the reading the UV cutoff gpio does? I don't see it in the schematic. Let's talk at standup

projects/uv_cutoff/src/main.c
```diff
@@ -0,0 +1,119 @@
...
+    return;
+  }
+
+  if (gpio_get_state(&uv_status, &uv_value) != STATUS_CODE_OK) {
+    LOG_CRITICAL("Error reading UV_cutoff pin!\n");
+    return;
+  }
+  if (uv_value == GPIO_STATE_LOW) {
+    is_disconnected = true;
+    set_uv_cutoff_notification_signal1(true);
```
mitchellostler, line 59: So is the only way we can recover out of this state resetting the controller board? seems like we would want to go back to a connected state if we read a GPIO high state?

devAdhiraj: Ah yes, I wasn't sure on whether we can recover like this or not. I guess it would make sense to back to connected if GPIO went HIGH again. Should I implement this?

mitchellostler: Yeah, I think that would be a good idea

devAdhiraj: Hm. Sure. So then, should I also change is_disconnected state if it pin goes back to HIGH?

mitchellostler (approved):

PR #151, opened 2023-02-14
Smoke I2C FWXV-189

ay0fasan:


mitchellostler (approved):
Looks good! I like how you setup the test with the defines, good thinking

PR #152, opened 2023-02-15
Fwxv 150 implement lights fsm

Citrusboa:


projects/power_distribution/src/lights_fsm.c
```diff
@@ -0,0 +1,147 @@
...
+    fsm_transition(fsm, HAZARD);
+  } else if ((light_event == EE_LIGHT_TYPE_SIGNAL_LEFT) && (light_state == EE_LIGHT_STATE_ON)) {
+    fsm_transition(fsm, LEFT_SIGNAL);
+  } else if ((light_event == EE_LIGHT_TYPE_SIGNAL_RIGHT) && (light_state == EE_LIGHT_STATE_ON)) {
+    fsm_transition(fsm, RIGHT_SIGNAL);
+  }
+}
+
+static void prv_init_state_output(void *context) {
+  LOG_DEBUG("Transitioned to INIT_STATE");
```
mitchellostler, line 62: We should set light_id_callback to NUM_EE_LIGHT_TYPE here, so softtimer doesn't retrigger

mitchellostler (changes requested):
Looks good!

the main things are: 
- I am realizing bps fault needs to be in its own module, since we still want the lights to work if we have a bps fault. 
- All states should be able to be transferred between, based on the input. If hazard is pressed, then it should override everything, until it is pressed a second time. Then it will need to return to whichever state it was in previously. You can store the previous state in the context pointer which is passed to each function, and set it in the input right before you transition.

projects/power_distribution/src/lights_fsm.c
```diff
@@ -0,0 +1,136 @@
...
+static void prv_lights_signal_blinker(SoftTimerId id) {
+  if (light_id_callback == EE_LIGHT_TYPE_SIGNAL_LEFT) {
+    gpio_toggle_state(&LEFT_LIGHT_ADDR);
+    soft_timer_start(600, prv_lights_signal_blinker, &s_timer_single);
+
+  } else if (light_id_callback == EE_LIGHT_TYPE_SIGNAL_RIGHT) {
+    gpio_toggle_state(&RIGHT_LIGHT_ADDR);
+    soft_timer_start(600, prv_lights_signal_blinker, &s_timer_single);
+
+  } else if (light_id_callback == EE_LIGHT_TYPE_SIGNAL_HAZARD) {
```
mitchellostler, line 24: Should set right state = left state so that the lights are on/off at the same time

projects/power_distribution/src/lights_fsm.c
```diff
@@ -0,0 +1,136 @@
...
+// Softtimer module setup for light blinkers
+static SoftTimer s_timer_single;
+static EELightType light_id_callback;
+
+FSM(lights, NUM_LIGHTS_STATES);
+
+static void prv_lights_signal_blinker(SoftTimerId id) {
+  if (light_id_callback == EE_LIGHT_TYPE_SIGNAL_LEFT) {
+    gpio_toggle_state(&LEFT_LIGHT_ADDR);
+    soft_timer_start(600, prv_lights_signal_blinker, &s_timer_single);
```
mitchellostler, line 18: Can we have 600 as a define (SIGNAL_BLINK_PERIOD_MS)

smoke/uart/config.json
mitchellostler, line 1: Is there a reason you removed these?

projects/power_distribution/src/lights_fsm.c
```diff
@@ -0,0 +1,136 @@
...
+}
+
+static void prv_init_state_output(void *context) {
+  LOG_DEBUG("Transitioned to INIT_STATE");
+  light_id_callback = NUM_EE_LIGHT_TYPES;
+}
+
+static void prv_left_signal_input(Fsm *fsm, void *context) {
+  // can transition to INIT, RIGHT, HAZARD
+  EELightType light_event = get_steering_info_analog_input();
```
mitchellostler, line 54: The HAZARD event will actually come from a centre console message, since that is where the button is. You will need to add this message to Centre console, and get it separately to the steering message. 

I added the definition to the [can message spec](https://uwmidsun.atlassian.net/wiki/spaces/ELEC/pages/3222438052/CAN+Message+Spec+-+MSXV)
just add an entry to the centre console can yaml, and then you'll be able to get the hazard state (1 hazard on, 0 hazard off).

mitchellostler (commented):
You're really close! Just a couple more comments.

projects/power_distribution/src/lights_fsm.c
```diff
@@ -0,0 +1,141 @@
...
+    gpio_toggle_state(&RIGHT_LIGHT_ADDR);
+    soft_timer_start(SIGNAL_BLINK_PERIOD_MS, prv_lights_signal_blinker, &s_timer_single);
+
+  } else if (light_id_callback == EE_LIGHT_TYPE_SIGNAL_HAZARD) {
+    GpioState sync_state = gpio_get_state(&LEFT_LIGHT_ADDR, &sync_state);
+    gpio_set_state(&RIGHT_LIGHT_ADDR, sync_state);
+    gpio_toggle_state(&LEFT_LIGHT_ADDR);
+    gpio_toggle_state(&RIGHT_LIGHT_ADDR);
+    soft_timer_start(SIGNAL_BLINK_PERIOD_MS, prv_lights_signal_blinker, &s_timer_single);
+  } else {
```
mitchellostler, line 30: Can you set the GPIOs off here (since we're transitioning to off state

projects/power_distribution/src/lights_fsm.c
```diff
@@ -0,0 +1,141 @@
...
+  } else if (light_event == EE_LIGHT_TYPE_SIGNAL_RIGHT) {
+    fsm_transition(fsm, RIGHT_SIGNAL);
+  }
+}
+
+static void prv_left_signal_output(void *context) {
+  LOG_DEBUG("Transitioned to LEFT_SIGNAL");
+  // Toggle Left Signal blinkers at 100 BPM -> 0.6s
+  light_id_callback = EE_LIGHT_TYPE_SIGNAL_LEFT;
+  soft_timer_start(600, prv_lights_signal_blinker, &s_timer_single);
```
mitchellostler, line 73: Can you add a define for all of these 600 ms params:
ie `BLINKER_PERIOD_MS`

mitchellostler (approved):

PR #155, opened 2023-02-19
Fwxv 188 smoke uart

kevinksong:


projects/smoke_uart/src/main.c
```diff
@@ -0,0 +1,62 @@
+#include <stdio.h>
+
+#include "interrupt.h"
+#include "log.h"
+#include "tasks.h"
+#include "uart.h"
+
+#define SHOULD_TRANSMIT true
```
JarvisWeng, line 8: What's the purpose of these macros

projects/smoke_uart/src/main.c
```diff
@@ -0,0 +1,62 @@
...
+    if (SHOULD_TRANSMIT) {
+      status = uart_tx(UART_PORT_SEND, &data_sent, &len);
+      if (status == STATUS_CODE_OK) {
+        LOG_DEBUG("Successfully transmitted %u\n", data_sent);
+      } else {
+        LOG_DEBUG("Write failed: status code %d\n", status);
+      }
+    }
+    if (SHOULD_READ) {
+      status = uart_rx(UART_PORT_RECEIVE, &data_received, &len);
```
JarvisWeng, line 40: Hm not too sure if this is what you need @mitchellostler

projects/smoke_uart/src/main.c
kevinksong: I was looking at some older smoke tests and they all had similar macros so I included them here aswell

projects/smoke_uart/src/main.c
mitchellostler: @kevinksong Yeah, I think you don't need the should_send/should_read for this. If you could just change it so there's a static char  buffer at the top with a string in it  (ie `static const char[] = "test uart\n" , and then just run the tx every second.  we don't need to do rx right now, since we don't use it for anything at the moment.

kevinksong:
I made the changes as requested. Please take a look and see if they're correct

projects/smoke_uart/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+                                      .baudrate = 9600 };
+
+TASK(master_task, TASK_MIN_STACK_SIZE) {
+  uart_init(UART_PORT_1, &uart_settings);
+
+  size_t len = sizeof(uint8_t);
+
+  while (true) {
+    StatusCode status;
+    status = uart_tx(UART_PORT_1, test, &len);
```
mitchellostler, line 22: len should be sizeof(test), so we transmit the whole string

mitchellostler (commented):
One more thing, otherwise looks good!

mitchellostler (approved):

PR #156, opened 2023-02-20
7seg driver

A3sanjay:


libraries/codegen/templates/_rx_structs.h.jinja
```diff
@@ -4,7 +4,7 @@
 {% set messages = data["Messages"] -%}
 #include <stdint.h>
 
-typedef struct
+typedef g
```
JarvisWeng, line 7: Why did you change this?

libraries/seg_display_driver/inc/seg_display.h
```diff
@@ -0,0 +1,53 @@
+#include "gpio.h"
+#include "gpio_it.h"
+typedef struct PinsBcd {
+  const GpioAddress A = {
```
JarvisWeng, line 4: What are you doing here?
You're `declaring` a struct. So you need to define the stuff that will be inside the struct. These are things like `uint8_t`, `bool`, `int32_t`, etc

PR #160, opened 2023-03-10
Steering analog test

A3sanjay:


libraries/ms-common/inc/adc.h
```diff
@@ -65,3 +65,9 @@ StatusCode adc_read_converted(GpioAddress address, uint16_t *reading);
 
 // Used to clear state
 void adc_deinit(void);
+
+// Used to set and read values modified from ADC pins
+uint16_t reading;
```
JarvisWeng, line 70: You shouldn't create variables in header files.
Unless it's prefixed with `extern`, ever file that includes this `adc.h` will create a global variable.
`reading` should be in `adc.c` and also `static`

Name it as so
`static uint16_t s_test_reading;`

JarvisWeng: Also what is this variable even used for?

libraries/ms-common/src/x86/adc.c
```diff
@@ -233,3 +233,13 @@ StatusCode adc_read_converted(GpioAddress address, uint16_t *reading) {
   *reading = (adc_reading * vdda) / 4095;
   return STATUS_CODE_OK;
 }
+
+// Getters and setters to test if ADC works
+void set_reading(GpioAddress sample_address) {
+  adc_add_channel(sample_address);
+  adc_init(ADC_MODE_SINGLE);
+  adc_read_converted(sample_address, &reading);
```
JarvisWeng, line 241: Hm I don't really think this is what @mitchellostler intended.
What should happen is that you edit the `s_adc_stores[channel].reading`  so that `adc_mock` will be able to change the reading in there.

projects/steering/test/test_steering_analog.c
```diff
@@ -0,0 +1,60 @@
...
+  set_reading(s_ctrl_stk_address);
+  control_stalk_data_verify = get_reading();
+  steering_analog_input();
+  while (true) {
+  }
+}
+
+TEST_IN_TASK
+void test_steering_analog(void) {
+  adc_init(ADC_MODE_SINGLE);
```
JarvisWeng, line 38: Hm not too sure what this is supposed to test. @mitchellostler are you testing CAN messages going into `steering`?

libraries/ms-common/src/x86/adc.c
```diff
@@ -233,3 +230,15 @@ StatusCode adc_read_converted(GpioAddress address, uint16_t *reading) {
   *reading = (adc_reading * vdda) / 4095;
   return STATUS_CODE_OK;
 }
+
+#ifdef MS_TEST
```
mitchellostler, line 234: You don't need the ifdef here

projects/steering/src/steering_analog.c
```diff
@@ -12,6 +12,7 @@ StatusCode steering_analog_adc_init() {
 
 void steering_analog_input() {
   uint16_t control_stalk_data;
+  uint8_t channel;
```
mitchellostler, line 15: Is there a reason for this?

PR #163, opened 2023-03-13
Fwxv 167 can watchdog

JarvisWeng:


can/src/can.c
```diff
@@ -68,7 +69,12 @@ TASK(CAN_TX, TASK_MIN_STACK_SIZE)
     LOG_DEBUG("can_tx called: %d!\n", counter);
     counter++;
 
+    check_can_watchdogs();
+
     can_tx_all();
+    
+    clear_rx_struct();
+    clear_tx_struct();
```
ShiCheng-Lu, line 77: I don't think you should clear data here. expose the function but don't always clear it.

ShiCheng-Lu: if you need to clear the received flag, probably best to autogen the setter for that

libraries/codegen/templates/_rx_all.c.jinja
```diff
@@ -3,6 +3,13 @@
 
 #include "can_board_ids.h"
 #include "can_codegen.h"
+#include "can_watchdog.h"
+
+{% for message, message_data in messages.items() %}
+    {%- if message_data["target"][board]["watchdog"] != 0 %}
+static CanWatchDog s_{{message}}_watchdog = {0, {{message_data["target"][board]["watchdog"] | lower}}};
```
ShiCheng-Lu, line 10: can you change this to s_{{message}}_msg_watchdog?

libraries/codegen/boards/new_can.yaml
```diff
@@ -3,8 +3,10 @@
     transmit_msg1:
       id: 0
       target:
-        - centre_console
-        - new_can
+        centre_console:
+          watchdog: 3
+        new_can:
+          watchdog: True
```
ShiCheng-Lu, line 9: why are some of these True/false, some numbers? (this + libraries/codegen/boards/babydriver.yaml)

libraries/codegen/templates/_rx_all.c.jinja
```diff
@@ -19,10 +26,34 @@ void can_rx_all() {
...
+    {%- if message_data["target"][board]["watchdog"] != 0 %}
+    if (!g_rx_struct.received_{{message}}) {
+        ++s_{{message}}_watchdog.cycles_over;
+        if (s_{{message}}_watchdog.cycles_over >= s_{{message}}_watchdog.max_cycles) {
+#ifdef MS_PLATFORM_X86
+        LOG_CRITICAL("DID NOT RECEIVE CAN MESSAGE: %u IN MAX CYCLES : %u\n", SYSTEM_CAN_MESSAGE_{{message_data["src"] | upper}}_{{message | upper}},
+                        s_{{message}}_watchdog.max_cycles);
+#else
+        LOG_CRITICAL("DID NOT RECEIVE CAN MESSAGE: %u IN MAX CYCLES : %u\n", SYSTEM_CAN_MESSAGE_{{message_data["src"] | upper}}_{{message | upper}},
+                        s_{{message}}_watchdog.max_cycles);
```
ShiCheng-Lu, line 52: these two are exactly the same...

libraries/master/inc/master_task.h
```diff
@@ -0,0 +1,18 @@
...
+void set_master_cycle_time(uint32_t time_ms);
+void set_medium_cycle_count(uint32_t cycles);
+void set_slow_cycle_count(uint32_t cycles);
+
+void run_fast_cycle();
+void run_medium_cycle();
+void run_slow_cycle();
+
+StatusCode init_master_task();
+Task *get_master_task();
```
ShiCheng-Lu, line 14: why not use DECLARE_TASK(); here in the header instead of this get_master_task func?

libraries/master/src/master_task.c
```diff
@@ -0,0 +1,77 @@
...
+    vTaskEndScheduler();
+  }
+}
+
+TASK(master_task, TASK_MIN_STACK_SIZE) {
+  TickType_t xLastWakeTime = xTaskGetTickCount();
+  uint32_t counter = 1;
+  while (true) {
+    // LOG_DEBUG("counter: %u\n", counter);
+#ifdef TEST
```
ShiCheng-Lu, line 52: is this suppose to be MS_TEST?

ShiCheng-Lu (approved):
lgtm

libraries/master/inc/master_task.h
mitchellostler: +1 on this

libraries/master/src/master_task.c
```diff
@@ -0,0 +1,72 @@
...
+  if (delay != pdTRUE) {
+    ++s_cycles_over;
+    LOG_WARN("Master Task out of sync!! Currently over by %u cycles\n", s_cycles_over);
+  } else {
+    if (s_cycles_over != 0) {
+      --s_cycles_over;
+    }
+  }
+  if (s_cycles_over > MAX_CYCLES_OVER) {
+    vTaskEndScheduler();
```
mitchellostler, line 43: I think this function is not supported on hardware correct? 
Or does it just infinite loop?

We should maybe broadcast a fatal error message too

libraries/codegen/templates/_rx_all.c.jinja
```diff
@@ -19,10 +26,35 @@ void can_rx_all() {
...
+    g_rx_struct.received_{{message}} = false;
+{%- endfor %}
+}
+
+StatusCode check_can_watchdogs() {
+{%- for message, message_data in messages.items() %}
+    {%- if message_data["target"][board]["watchdog"] != 0 %}
+    if (!g_rx_struct.received_{{message}}) {
+        ++s_{{message}}_msg_watchdog.cycles_over;
+        if (s_{{message}}_msg_watchdog.cycles_over >= s_{{message}}_msg_watchdog.max_cycles) {
```
mitchellostler, line 52: We're going to need to do something other than log critical here, since we'll need a way of detecting which watchdog has failed so that we can act on it.

mitchellostler: We also never reset the watchdog count here if we do receive a message, and we should

libraries/codegen/templates/_rx_all.c.jinja
```diff
@@ -19,10 +26,35 @@ void can_rx_all() {
                 g_rx_struct.{{message}}_{{signal}} = (msg.data >> {{offset.value}});
                 {%- set offset.value = offset.value + signal_data.length %}
             {%- endfor %}
+                g_rx_struct.received_{{message}} = true;
```
mitchellostler, line 29: We should do this in can hardware no? Right now it looks like we're just setting it every cycle instead of when we actually have a new message that we've pulled of the CAN bus

mitchellostler (changes requested):

libraries/codegen/templates/_getters.h.jinja
```diff
@@ -10,4 +10,6 @@
 #define get_{{message}}_{{signal}}() \
     g_rx_struct.{{message}}_{{signal}}
     {% endfor %}
-{%- endfor %}
+#define get_recevied_{{message}} \
```
mitchellostler, line 13: Little typo here

mitchellostler (approved):
All looks good. 
Do we reset the received_{message} to false though? would make sense to do that in the tx_all I think.

PR #164, opened 2023-03-15
FWXV-195 Pedal Logic Test

ay0fasan:


projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,99 @@
...
+
+// Define a static variable to hold the sent CAN message
+static CanMessage sent_msg;
+
+// Define static variables to hold the simulated throttle and brake readings
+static int16_t simulated_throttle_reading = THROTTLE_POSITION;
+static int16_t simulated_brake_reading = BRAKE_POSITION;
+
+// Mock CAN transmit function to capture the transmitted message
+StatusCode TEST_MOCK(can_transmit)(const CanMessage *msg) {
```
Bafran, line 60: You won't need to mock can_transmit either, let me know if you need the CAN documentation, it'll make it more clear how you should use it.

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,99 @@
...
+  sent_msg = *msg;
+  return STATUS_CODE_OK;
+}
+
+// Mock function for reading throttle data from MAX11600
+StatusCode TEST_MOCK(read_throttle_data)(int16_t *position) {
```
Bafran, line 66: I think we should mock the MAX_11600 function instead, and implement the read_throttle and read_brake which calls the MAX function. For the purpose of this ticket, you should also mock the pedal's calibration. Basically, the functionality in `pedal_data.c` should be verified too.

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,99 @@
...
+}
+
+void setup_test(void) {
+  log_init();
+  can_init(&can_storage, &can_settings);
+}
+
+void teardown_test(void) {}
+
+void test_pedal_can_tx(void) {
```
Bafran, line 84: Let's call this test_pedal_cycle so it's a bit more clear. The flow should be testing reading the pedal data, and after that additionally setting the CAN transmit value.

Bafran (changes requested):
I left a few comments, but essentially I think we should go about this by mocking the IC that's giving us our pedal readings, the calibration, and then using the CAN setters to set the values for us. Then at the end we'll assert that the correct values are in the struct. I might not be explaining this clearly, let me know if you have questions!

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
...
+#include "pedal_data.h"
+#include "pedal_calib.h"
+#include "soft_timer.h"
+#include "unity.h"
+
+#define THROTTLE_CHANNEL MAX11600_CHANNEL_0
+#define BRAKE_CHANNEL MAX11600_CHANNEL_2
+#define EE_PEDAL_VALUE_DENOMINATOR (1 << 12)
+
+extern g_tx_struct;
```
Bafran, line 22: You won't actually need to extern this variable, juts include `#include "pedal_setters.h"`, that file has the macros you used below and will also give you access to this variable.

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
...
+
+  //Read throttle and brake values
+  StatusCode status_throttle = read_pedal_data(&throttle_reading, THROTTLE_CHANNEL);
+  StatusCode status_brake = read_pedal_data(&brake_reading, BRAKE_CHANNEL);
+
+  TEST_ASSERT_OK(status_throttle);
+  TEST_ASSERT_OK(status_brake);
+
+  //Use CAN setters to set throttle and brake values
+  set_pedal_output_throttle_output(throttle_reading);
```
Bafran, line 71: These macros are included in the header I mentioned above

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
...
+  set_pedal_output_brake_output(brake_reading);
+
+  //Transmit the data over CAN
+  can_tx_all();
+
+  //Check the correct values for the throttle and brake signals are in the g_tx_struct
+  int32_t throttle_sent = (int32_t)(g_tx_struct.pedal_output.throttle_output);
+  int32_t brake_sent = (int32_t)(g_tx_struct.pedal_output.brake_output);
+  TEST_ASSERT_EQUAL_INT32(throttle_reading, throttle_sent);
```
Bafran, line 80: Here, let's check the reading in the CAN struct (throttle_sent  & brake_sent) against a calculated expected value, instead of just the value we set it to. These will be #define constants defined at the top of the file. You'll have to just run the number and see what the expected value is once it goes through the function. This is something that may change as we update pedal but if that ends up happening we'll just update the #define. Just make sure you can run the test and have it pass when you run `scons test --platform=x86 --project=pedal --test=pedal_data_tx`.

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
...
+
+#define THROTTLE_CHANNEL MAX11600_CHANNEL_0
+#define BRAKE_CHANNEL MAX11600_CHANNEL_2
+#define EE_PEDAL_VALUE_DENOMINATOR (1 << 12)
+
+extern g_tx_struct;
+
+//Mock function for pedal calibration
+PedalCalibBlob global_calib_blob_mock = {
+  .throttle_calib = { .lower_value = 0, .upper_value = 1000 },
```
Bafran, line 26: These numbers can be in a #define at the top of the file.

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
...
+  .brake_calib = { .lower_value = 0, .upper_value = 1000 },
+};
+
+//Mock function for max11600_read_converted
+StatusCode TEST_MOCK(max11600_read_converted)(Max11600Storage *storage) {
+  storage->channel_readings[THROTTLE_CHANNEL] = 500; //Mock throttle value
```
Bafran, line 32: Same for these numbers

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
+#include <stdint.h>
+#include "max11600.h"
+#include "can_hw.h"
+#include "can_msg.h"
+#include "can.h"
+#include "delay.h"
+#include "fsm.h"
+#include "gpio.h"
+#include "gpio_it.h"
```
Bafran, line 9: nit: I think a few of these don't need to be included, it's not really a big deal though

projects/pedal/test/test_pedal_data_tx.c
```diff
@@ -0,0 +1,82 @@
...
+
+  //Use CAN setters to set throttle and brake values
+  set_pedal_output_throttle_output(throttle_reading);
+  set_pedal_output_brake_output(brake_reading);
+
+  //Transmit the data over CAN
+  can_tx_all();
+
+  //Check the correct values for the throttle and brake signals are in the g_tx_struct
+  int32_t throttle_sent = (int32_t)(g_tx_struct.pedal_output.throttle_output);
```
Bafran, line 78: nit: this doesn't need to go in a variable, you can just assert directly against the contents of the tx struct

Bafran (changes requested):
This looks really good! Most of it is working but just a few changes, most importantly make the test run against a pre-calculated expected value. Before pushing your changes, run `scons format` and `scons lint` and make any changes it says to so that the CI will pass, and also run the test and see if it passes

mitchellostler:
@Bafran @ay0fasan Just checking in with this PR. What's the state of this? Does more need to be done?

PR #165, opened 2023-03-16
FWXV-196: pca9555 tests

devAdhiraj:
Added a couple of tests to show the general structure I'm following. I plan to add more.

I added a queue in the /x86/i2c library. This is based off the implementation in the arm version.

The issue with testing currently is that when reading/writing a register from i2c, it first writes the reg address and then reads from i2c. So if we use a dummy queue or a dummy variable, it will write the reg address to it, and then we will be reading that dummy value instead of any other pre-set value.

### DIFF
--- a/libraries/ms-common/src/x86/i2c.c
+++ b/libraries/ms-common/src/x86/i2c.c
@@ -3,12 +3,45 @@
 #include "log.h"
 #include "stdio.h"
 
+// Used to protect I2C resources
+// All data tx'd and rx'd through queue
+// Transactions on each I2C port protected by Mutex
+typedef struct {
+  uint8_t buf[I2C_MAX_NUM_DATA];
+  Queue queue;
+  Mutex mutex;
+} I2CBuffer;
+
+typedef enum I2CMode {
+  I2C_MODE_TRANSMIT = 0,
+  I2C_MODE_RECEIVE,
+  NUM_I2C_MODES,
+} I2CMode;
+
+typedef struct {
+  I2CSettings settings;     // I2C Config
+  I2CBuffer i2c_buf;        // RTOS data structures
+  bool curr_mode;           // Mode used in current transxn
+  I2CAddress current_addr;  // Addr used in current transxn
+} I2CPortData;
+
+static I2CPortData s_port[NUM_I2C_PORTS] = {
+  [I2C_PORT_1] = {},
+  [I2C_PORT_2] = {},
+};
+
 StatusCode i2c_init(I2CPort i2c, const I2CSettings *settings) {
   if (i2c >= NUM_I2C_PORTS) {
     return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C port.");
   } else if (settings->speed >= NUM_I2C_SPEEDS) {
     return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid I2C speed.");
   }
+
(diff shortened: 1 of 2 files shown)


libraries/ms-common/src/x86/i2c.c
```diff
@@ -3,22 +3,63 @@
 #include "log.h"
 #include "stdio.h"
 
+// Used to protect I2C resources
+// All data tx'd and rx'd through queue
+// Transactions on each I2C port protected by Mutex
+typedef struct {
+  uint8_t buf[I2C_MAX_NUM_DATA];
+  Queue queue;
+  Mutex mutex;
```
Bafran, line 12: I don't remember, did we want to implement the mutex logic as well in x86? It's not implemented right now. @mitchellostler

mitchellostler: Yeah I think it's a good idea to add the mutex logic if you can

libraries/ms-drivers/test/test_pca9555_gpio_expander.c
```diff
@@ -0,0 +1,215 @@
...
+#include "gpio.h"
+#include "log.h"
+#include "pca9555_gpio_expander.h"
+#include "semaphore.h"
+#include "status.h"
+#include "task_test_helpers.h"
+#include "tasks.h"
+#include "unity.h"
+
+#define MOCK_I2C_ADDRESS 0b01010101
```
mitchellostler, line 14: Don't use binary, always use hex. It's much more compact and easier to read

libraries/ms-drivers/test/test_pca9555_gpio_expander.c
```diff
@@ -0,0 +1,215 @@
...
+void teardown_test(void) {}
+
+TEST_IN_TASK
+
+void test_gpio_set_state(void) {
+  StatusCode status;
+  uint8_t data = MOCK_REG;
+
+  // set reg to a mock state
+  status = i2c_set_data(i2c_port, &data, 1);
```
mitchellostler, line 45: I don't think you need to set data here right? pca9555_gpio_set_state will be changing the state by sending data, ie writing to the rx_queue, so you should just be able to read it back

libraries/ms-drivers/test/test_pca9555_gpio_expander.c
```diff
@@ -0,0 +1,215 @@
...
+  TEST_ASSERT_EQUAL(STATUS_CODE_OK, status);
+
+  // set gpio pin to HIGH
+  status = pca9555_gpio_set_state(&pca9555_address, PCA9555_GPIO_STATE_HIGH);
+  TEST_ASSERT_EQUAL(STATUS_CODE_OK, status);
+
+  // checking the output reg address. Reg address should be 0x01 (OUTPUT0) for pins 0-7
```
mitchellostler, line 52: Comments don't match test?

mitchellostler (commented):
Overall looks good, just a couple of comments

libraries/ms-drivers/test/test_pca9555_gpio_expander.c
devAdhiraj: So pca9555_gpio_set_state reads the reg content first. Then, it modifies the bit in the reg content based on which pin is being set and then it writes the register with the updated value. So it still reads the register first which is why I have to set the reg to a mock state here.

PR #166, opened 2023-03-16
FWXV-165: Implement BTS7004 Driver

03predo:
This implementation does not include support for using the enable pin through a IO expander, but we can update if needed.


mitchellostler (approved):
We'll be integrating this into the other driver so it will get pulled out of the repo, but I'll merge it so you get the commit credits. Good work!

PR #167, opened 2023-03-18
Fwxv 191 create steering digital tests

Citrusboa:


mitchellostler (approved):

projects/steering/test/test_steering_digital.c
```diff
@@ -0,0 +1,102 @@
...
+void test_steering_input(void) {
+  // Empty notification
+  TEST_ASSERT_EQUAL(notification, 0);
+  TEST_ASSERT_EQUAL(steering_event, INVALID_EVENT);
+}
+
+TEST_IN_TASK
+void test_steering_digital_input_horn() {
+  // Test horn event & CAN message - press and unpress
+  TEST_ASSERT_OK(gpio_it_trigger_interrupt(&(GpioAddress)HORN_GPIO_ADDR));
```
mitchellostler, line 55: Yeah, you can't actually do this cast. when you do a compound literal, you have to assign it to a pointer and the compiler can figure out to allocate the memory for it. 

Just do GpioAddress horn = HORN_GPIO_ADDR at the top, and take the address of that. This will be the same for all of them. and that's why your build is failing.

PR #168, opened 2023-03-18
finished max11600 test

Scrubpai:


projects/pedal/test/test_pedal_max11600.c
```diff
@@ -0,0 +1,66 @@
+#include "log.h"
+#include "max11600.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+#define TEST_MAX11600_I2C_PORT I2C_PORT_2
+#define THROTTLE_CHANNEL MAX11600_CHANNEL_0
```
Bafran, line 7: Are these channels definitions needed?

projects/pedal/test/test_pedal_max11600.c
```diff
@@ -0,0 +1,66 @@
...
+  TEST_ASSERT_EQUAL(TEST_MAX11600_I2C_PORT, max_storage.i2c_port);
+  TEST_ASSERT_EQUAL(MAX11600_READ_ADDRESS, max_storage.i2c_read_address);
+  TEST_ASSERT_EQUAL(MAX11600_WRITE_ADDRESS, max_storage.i2c_write_address);
+  for (int channel = 0; channel < NUM_MAX11600_CHANNELS; channel++) {
+    TEST_ASSERT_EQUAL(0, max_storage.channel_readings[channel]);
+  }
+}
+
+void test_max11600_read_raw(void) {
+  for (int LSB_THROTTLE = 0; LSB_THROTTLE < 256; LSB_THROTTLE++) {
```
Bafran, line 31: We don't necessarily need to mock every single possibility, let's just try out maybe 5 values between 0 - 3.3V

projects/pedal/test/test_pedal_max11600.c
```diff
@@ -0,0 +1,66 @@
...
+    for (int LSB_STEERING = 0; LSB_STEERING < 256; LSB_STEERING++) {
+      for (int LSB_BRAKE = 0; LSB_BRAKE < 256; LSB_BRAKE++) {
+        // Mock all possible combinations of LSD onto the I2C
+        uint8_t tx_data[] = { LSB_THROTTLE, LSB_STEERING, LSB_BRAKE };
+        i2c_write(TEST_MAX11600_I2C_PORT, MAX11600_WRITE_ADDRESS, tx_data, 3);
```
Bafran, line 36: This should be changing soon, we'll have some kind of I2C setter function so we can set a value that will then be read when the i2c_read() function is called.

projects/pedal/test/test_pedal_max11600.c
```diff
@@ -0,0 +1,66 @@
...
+        for (int channel = 0; channel < 3; channel++) {
+          TEST_ASSERT_EQUAL(tx_data[channel], max_storage.channel_readings[channel]);
+        }
+      }
+    }
+  }
+}
+
+void test_max11600_read_converted(void) {
+  for (int LSB_THROTTLE = 0; LSB_THROTTLE < 256; LSB_THROTTLE++) {
```
Bafran, line 48: Same thing here, we don't need to mock every value.

projects/pedal/test/test_pedal_max11600.c
```diff
@@ -0,0 +1,66 @@
...
+      for (int LSB_BRAKE = 0; LSB_BRAKE < 256; LSB_BRAKE++) {
+        // Mock all possible combinations of LSD onto the I2C
+        uint8_t tx_data[] = { LSB_THROTTLE, LSB_STEERING, LSB_BRAKE };
+        i2c_write(TEST_MAX11600_I2C_PORT, MAX11600_WRITE_ADDRESS, tx_data, 3);
+        // Read the converted data from the MAX11600
+        TEST_ASSERT_OK(max11600_read_converted(&max_storage));
+        for (int channel = 0; channel < 3; channel++) {
+          // Calculate the expected converted value
+          uint16_t expected_converted_value = 1000 * REFERENCE_VOLTAGE_V * tx_data[channel] / 256;
+          TEST_ASSERT_EQUAL(expected_converted_value, max_storage.channel_readings[channel]);
```
Bafran, line 59: Try run these tests, I ran it on my computer and the asserts were failing for I think both tests, so just make sure that it is actually working. It's also possible that it is in fact the MAX driver failing, in which case we should fix that bug.

Bafran (changes requested):

projects/pedal/src/max11600.c
```diff
@@ -25,6 +25,11 @@
 
 static const uint8_t i2c_read_address = MAX11600_READ_ADDRESS;
 static const uint8_t i2c_write_address = MAX11600_WRITE_ADDRESS;
+static const I2CSettings i2c_settings = {
```
mitchellostler, line 28: This shouldn't be in the driver file, this should be passed from main to max11600 init. Also the driver should be in the ms-drivers library. @Bafran did you fix this on master or was that something else?

projects/pedal/test/test_pedal_max11600.c
```diff
@@ -0,0 +1,116 @@
+#include "log.h"
+#include "max11600.h"
+#include "test_helpers.h"
+#include "unity.h"
+
+#define TEST_MAX11600_I2C_PORT I2C_PORT_2
+#define REFERENCE_VOLTAGE_V 3.3
```
mitchellostler, line 7: using decimals in a defined macro can create all sorts of problems, change this to mV please. It won't be a problem with how you used it, but is really not good practice

mitchellostler (commented):
Just the two things, otherwise looks good!

PR #172, opened 2023-03-22
Fwxv 201 motor controller pre charge signal

kevinksong:
Couple Notes:
1. In main.c for the PrechargeControlSettings I initialized the gpio ports randomly, not sure what ports will actually be used
2. In precharge_control_init, I initialized the tasks PRECHARGE_INTERRUPT and PRECHARGE_INTERRUPT2 with task priority 1, not sure if this is correct or not


projects/motor_controller/src/precharge_control.c
```diff
@@ -0,0 +1,74 @@
...
+      } else {
+        set_mc_status_precharge_status(MCI_PRECHARGE_CHARGED);
+      }
+    }
+  }
+}
+
+StatusCode run_precharge_rx_cycle() {
+  if (get_drive_output_precharge() == MCI_PRECHARGE_CHARGED) {
+    set_mc_status_precharge_status(MCI_PRECHARGE_CHARGED);
```
ShiCheng-Lu, line 49: shouldn't set this status here, the status only depend on the gpio interrupts

projects/motor_controller/inc/precharge_control.h
```diff
@@ -0,0 +1,38 @@
...
+  GpioAddress precharge_control;
+  GpioAddress precharge_monitor;
+  GpioAddress precharge_monitor2;
+} PrechargeControlSettings;
+
+typedef struct PrechargeControlStorage {
+  GpioAddress precharge_control;
+} PrechargeControlStorage;
+
+PrechargeControlStorage *test_get_storage(void);
```
ShiCheng-Lu, line 27: if this is only for testing, wrap it in #ifdef MS_TEST, so that it will only be compiled when running the tests

projects/motor_controller/inc/precharge_control.h
```diff
@@ -0,0 +1,38 @@
...
+
+typedef struct PrechargeControlSettings {
+  GpioAddress precharge_control;
+  GpioAddress precharge_monitor;
+  GpioAddress precharge_monitor2;
+} PrechargeControlSettings;
+
+typedef struct PrechargeControlStorage {
+  GpioAddress precharge_control;
+} PrechargeControlStorage;
```
ShiCheng-Lu, line 25: you can just have a static GpioAddress instead of this struct since this struct only holds one item

projects/motor_controller/src/precharge_control.c
```diff
@@ -0,0 +1,74 @@
...
+                          const PrechargeControlSettings *settings) {
+  storage->precharge_control = settings->precharge_control;
+}
+
+StatusCode precharge_control_init(const PrechargeControlSettings *settings) {
+  interrupt_init();
+  prv_populate_storage(&s_precharge_storage, settings);
+  InterruptSettings monitor_it_settings = { .type = INTERRUPT_TYPE_INTERRUPT,
+                                            .priority = INTERRUPT_PRIORITY_NORMAL,
+                                            .edge = INTERRUPT_EDGE_RISING };
```
ShiCheng-Lu, line 67: nit: add a comma at the end here so the format is better.

projects/motor_controller/src/precharge_control.c
```diff
@@ -0,0 +1,74 @@
...
+  } else {
+    set_mc_status_precharge_status(MCI_PRECHARGE_DISCHARGED);
+    return prv_set_precharge_control(&s_precharge_storage, GPIO_STATE_LOW);
+  }
+}
+
+void prv_populate_storage(PrechargeControlStorage *storage,
+                          const PrechargeControlSettings *settings) {
+  storage->precharge_control = settings->precharge_control;
+}
```
ShiCheng-Lu, line 60: inline this function as it's not doing much

projects/motor_controller/inc/precharge_control.h
ShiCheng-Lu: also wrap the implementation

projects/motor_controller/inc/precharge_control.h
```diff
@@ -0,0 +1,29 @@
...
+  GpioAddress precharge_control;
+  GpioAddress precharge_monitor;
+  GpioAddress precharge_monitor2;
+} PrechargeControlSettings;
+
+StatusCode prv_set_precharge_control(GpioAddress *address, const GpioState state);
+
+StatusCode run_precharge_rx_cycle();
+
+PrechargeState get_precharge_state();
```
ShiCheng-Lu, line 27: not used and not implemented?

ShiCheng-Lu (approved):

PR #175, opened 2023-03-25
FWXV-193 CAN debug module

elbertchen1:


elbertchen1:
@JarvisWeng I moved the library files to the can directory because there were some implicit dependency errors when I had it in ms-common

can/src/can_debug.c
```diff
@@ -0,0 +1,35 @@
+#include "can_debug.h"
+#include "log.h"
+
+static int i;
```
JarvisWeng, line 4: give this a more meaningful name. Like `s_num_can_callback_functions` or something

can/src/can_debug.c
```diff
@@ -0,0 +1,35 @@
...
+static CanMessage msg;
```
JarvisWeng, line 5: Isn't `msg` just used in the function? It doesn't need to global?

can/src/can_debug.c
```diff
@@ -0,0 +1,35 @@
...
+    i++;
+    return STATUS_CODE_OK;
+  } else {
+    LOG_CRITICAL("The max number of callbacks has been reached!");
+    return STATUS_CODE_RESOURCE_EXHAUSTED;
+  }
+}
+
+void check_can_messages() {
+  if (can_receive(&msg) == STATUS_CODE_OK) {
```
JarvisWeng, line 28: Make this a while loop instead

can/src/can_debug.c
```diff
@@ -0,0 +1,35 @@
...
+    for (int i = 0; i < NUM_CALLBACKS; i++) {
```
JarvisWeng, line 29: this is why you don't name things the same. `i` makes it unclear if it's the global or local `i`. In the case, local usually has priority

projects/can_debug/src/main.c
```diff
@@ -1,66 +1,51 @@
...
 const CanSettings can_settings = {
   .device_id = 0x1,
   .bitrate = CAN_HW_BITRATE_500KBPS,
   .tx = { GPIO_PORT_A, 12 },
   .rx = { GPIO_PORT_A, 11 },
   .loopback = true,
+  .mode = 1,
 };
 
-int perform_calculation(uint8_t operation, uint8_t operandA, uint8_t operandB, uint8_t operandC) {
```
JarvisWeng, line 18: Oh can you also do the same operations as before?
Like if you get this message, do an add operation, etc.

can/inc/can_debug.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+// CAN Debug library
+//
+// Allows callback functions to be registered with specific CAN Message IDs.
+//
+// See fwxv/projects/can_debug for a reference project.
+
+#include "can.h"
```
JarvisWeng, line 8: I think you only need `can_msg.h`

can/src/can_debug.c
```diff
@@ -0,0 +1,34 @@
+#include "can_debug.h"
+#include "log.h"
+
+static int s_num_can_callback_functions;
```
JarvisWeng, line 4: maybe make this a `uint8_t`. Just so it saves space and it can't be negative

projects/can_debug/src/main.c
```diff
@@ -1,66 +1,73 @@
...
-int main(void) {
+int main() {
   tasks_init();
   log_init();
 
-  LOG_DEBUG("Welcome to CAN!\n");
+  LOG_DEBUG("Welcome to CAN Debug!\n");
   can_init(&s_can_storage, &can_settings);
 
+  CanMessageId add_id = 0x01E;
```
JarvisWeng, line 61: Nit: One more for just `Hello World`. Maybe `0x00E`

projects/can_debug/src/main.c
```diff
@@ -1,66 +1,73 @@
...
+#define LAST_BYTE_BIT_SHIFT 56
+
 static CanStorage s_can_storage = { 0 };
 const CanSettings can_settings = {
   .device_id = 0x1,
   .bitrate = CAN_HW_BITRATE_500KBPS,
   .tx = { GPIO_PORT_A, 12 },
   .rx = { GPIO_PORT_A, 11 },
   .loopback = true,
+  .mode = 1,
```
JarvisWeng, line 16: I think there's actually a `#define` for this instead of one. Is should be something like `CAN_ONESHOT_MODE` or something. Could you use that instead of `1`?

JarvisWeng (commented):
Looks good. Just a few more small things

JarvisWeng (approved):

PR #180, opened 2023-04-14
FWXV-209 Move pedal test to smoke directory

ay0fasan:


x86_flash
mitchellostler, line 1: Do we want this file?

PR #183, opened 2023-05-12
Refactor/Update Pedal

Bafran:


libraries/ms-drivers/src/max11600.c
```diff
@@ -25,11 +25,6 @@
 
 static const uint8_t i2c_read_address = MAX11600_READ_ADDRESS;
 static const uint8_t i2c_write_address = MAX11600_WRITE_ADDRESS;
-static const I2CSettings i2c_settings = {
```
JarvisWeng, line 28: Do you not need to initialize `i2c`?

libraries/ms-drivers/test/test_max11600.c
```diff
@@ -9,10 +10,16 @@
...
+static I2CSettings i2c_settings = {
+  .speed = I2C_SPEED_FAST,
+  .scl = { .port = GPIO_PORT_B, .pin = 10 },
+  .sda = { .port = GPIO_PORT_B, .pin = 11 },
+};
 
 void setup_test(void) {
   log_init();
   LOG_DEBUG("here");
+  i2c_init(TEST_MAX11600_I2C_PORT, &i2c_settings);
```
JarvisWeng, line 22: hm do you just need to port out the `i2c` outside the driver for tests?

JarvisWeng (approved):
Yeah looks fine to me. Other than the I2C moving, but I guess people will just need to remember to initialize it

PR #184, opened 2023-05-14
Centre console display

A3sanjay:


projects/centre_console/src/cc_displays.c
```diff
@@ -0,0 +1,35 @@
...
+#include "centre_console_getters.h"
+#include "centre_console_setters.h"
+#include "centre_console_tx_structs.h"
+#include "gpio.h"
+#include "log.h"
+#include "seg_display.h"
+
+void cc_displays_update(void) {
+  // Read data from CAN structs and update displays with those values
+  uint8_t cruise_control_val = g_tx_struct.drive_output_cruise_control;
```
JarvisWeng, line 14: Why do you need the `tx` value for this? Ideally, you should just pass where this is set.

projects/centre_console/src/cc_displays.c
```diff
@@ -0,0 +1,35 @@
...
+  // Convert motor_velocity value to speed by averaging the right and left, dividing by 100, and
+  // multiplying by 3.6 to get kph
+  float speed_val = (get_motor_velocity_velocity_l() + get_motor_velocity_velocity_r()) * 0.018;
```
JarvisWeng, line 17: Get rid of `0.018` and make that a #define. So you don't need have magic numbers

projects/centre_console/src/cc_displays.c
```diff
@@ -0,0 +1,35 @@
...
+  uint16_t batt_perc_val = get_battery_vt_batt_perc();
+  SegDisplay cc_display = { CC_DISP1, CC_DISP2, CC_DISP3, CC_DISP4,
+                            CC_DISP5, CC_DISP6, CC_DISP7, CC_DISP8 };
+  SegDisplay speed_display = { SPD_DISP1, SPD_DISP2, SPD_DISP3, SPD_DISP4,
+                               SPD_DISP5, SPD_DISP6, SPD_DISP7, SPD_DISP8 };
+  SegDisplay batt_perc_display = { BATT_DISP1, BATT_DISP2, BATT_DISP3, BATT_DISP4,
+                                   BATT_DISP5, BATT_DISP6, BATT_DISP7, BATT_DISP8 };
+  seg_display_init(&cc_display);
+  seg_display_init(&speed_display);
+  seg_display_init(&batt_perc_display);
```
JarvisWeng, line 27: Do these need to be re`init` each time this function is called? If not, maybe these could be global variables

can/src/can.c
```diff
@@ -1,12 +1,12 @@
```
mitchellostler, line 1: I'm a bit confused what these changes are for? Was the format failing?

libraries/ms-drivers/src/seg_display.c
```diff
@@ -48,7 +50,6 @@ StatusCode seg_display_set_float(SegDisplay *display, float val) {
   }
 }
 
-void set_seg_display(SegDisplay *display, uint8_t disp_value, uint8_t pin_number);
```
mitchellostler, line 51: Is there a reason why you're forward declaring the functions in the C file? If not then you can probably just remove the declaration

projects/centre_console/src/cc_displays.c
```diff
@@ -0,0 +1,40 @@
...
+#include "centre_console_setters.h"
+#include "centre_console_tx_structs.h"
+#include "gpio.h"
+#include "log.h"
+#include "seg_display.h"
+
+// Multiplication Factor to convert CAN Velocity in 100 * m/s to kph
+#define CONVERT_VELOCITY_TO_SPEED 0.018
+
+SegDisplay cc_display = { CC_DISP1, CC_DISP2, CC_DISP3, CC_DISP4,
```
mitchellostler, line 15: when you're initializing, can you specify each param, ie .A = CC_DISP1, .B = ... etc on new lines, it just makes it a bit less error prone

projects/centre_console/src/cc_displays.c
mitchellostler: I think this is fine for now since we don't have that cruise control value yet, but add a todo to get it right from the source.

mitchellostler (commented):
A few notes, otherwise looks good

PR #188, opened 2023-05-21
added bts load switch functionality except init function

harsh-patell:


mitchellostler (changes requested):
Overall looks really good, all the logic looks correct. I just added a couple of comments to make the interface a bit cleaner. Also, not sure what the comments are at the end of your header file?

mitchellostler (approved):

libraries/ms-drivers/inc/bts_load_switch.h
```diff
@@ -0,0 +1,63 @@
...
+// Holds pin-specific info for EN/SEL pins
+typedef struct {
+  GpioAddress *pin_stm32;
+  Pca9555GpioAddress *pin_pca9555;
+  Bts7xxxPinType pin_type;
+} Bts7xxxPin;
+
+// Holds info for the select pin state
+typedef struct {
+  GpioState *select_state_stm32;
```
mitchellostler, line 27: This should not be a pointer, it should just be a GpioState. This is what is causing your build failures

harsh-patell: I made the appropriate change, build is green now

mitchellostler (approved):

PR #192, opened 2023-05-24
fwxv_158_can_validation

mitchellostler:
- Removed filter out
- Added push from isr methods 
- Improved New can project to display sent/received data, verified on HW

### DIFF
--- a/can/src/can.c
+++ b/can/src/can.c
@@ -45,7 +45,7 @@ StatusCode run_can_tx_cycle()
   return STATUS_CODE_OK;
 }
 
-TASK(CAN_RX, TASK_MIN_STACK_SIZE)
+TASK(CAN_RX, 512)
 {
   int counter = 0;
   while (true)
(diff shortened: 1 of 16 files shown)


can/src/can.c
```diff
@@ -60,7 +60,7 @@ TASK(CAN_RX, TASK_MIN_STACK_SIZE)
   }
 }
 
-TASK(CAN_TX, TASK_MIN_STACK_SIZE)
+TASK(CAN_TX, 512)
```
JarvisWeng, line 63: Why make this a magic number?

libraries/ms-common/src/arm/uart.c
```diff
@@ -30,6 +30,7 @@ static UartPortData s_port[] = {
                     .periph = RCC_APB2Periph_USART1,
                     .irq = USART1_IRQn,
                     .base = USART1 },
+#ifdef MULTI_PORT_UART
```
JarvisWeng, line 33: Did you just do this to save space or was there an actual issue with multiple UART ports?

JarvisWeng (approved):

libraries/ms-common/src/arm/uart.c
mitchellostler: Nah just to save space. Multi-uart port hasn't been tested, but should be the exact same functionality. Testing it is a low priority since we don't need anything other than logs at the moment

PR #198, opened 2023-06-22
ADC Hardware Validation

mitchellostler:
Adc is working babyyyyy


libraries/ms-common/src/arm/adc.c
```diff
@@ -146,156 +117,150 @@ StatusCode adc_add_channel(GpioAddress address) {
...
-    .ADC_DataAlign = ADC_DataAlign_Right,  // Use rightmost 12 bits of ADC register
-    .ADC_NbrOfChannel = s_adc_status.active_channels,  // Scan as many channels as we've initialized
-  };
-  ADC_Init(ADC1, &adc_settings);
+  // By default, enable vref and temp sensor for voltage conversions
+  ADC_TempSensorVrefintCmd(true);
 
   // Enable the ADC
   ADC_Cmd(ADC1, true);
+  ADC_TempSensorVrefintCmd(ENABLE);
```
JarvisWeng, line 191: Is this a copy? Or why is `ADC_TempSensorVrefintCmd` called twice?

libraries/ms-common/src/arm/adc.c
```diff
@@ -146,156 +117,150 @@ StatusCode adc_add_channel(GpioAddress address) {
...
-  switch (channel) {
-    case ADC_Channel_TempSensor:
-      *reading = prv_get_temp(adc_reading);
-      return STATUS_CODE_OK;
-    case ADC_Channel_Vrefint:
-      *reading = prv_get_vdda(adc_reading);
-      return STATUS_CODE_OK;
-    default:
-      break;
+  if (channel == ADC_Channel_TempSensor) {
```
JarvisWeng, line 244: so only doing temperature for now?

libraries/ms-freertos/inc/FreeRTOSConfig.h
```diff
@@ -88,7 +88,7 @@ extern uint32_t SystemCoreClock;
 #define INCLUDE_vTaskDelay 1
 #define INCLUDE_xTaskGetSchedulerState 1
 #define INCLUDE_xTaskGetCurrentTaskHandle 1
-#define INCLUDE_uxTaskGetStackHighWaterMark 0
+#define INCLUDE_uxTaskGetStackHighWaterMark 1
```
JarvisWeng, line 91: What's this for?

libraries/ms-common/src/arm/adc.c
```diff
@@ -146,156 +117,150 @@ StatusCode adc_add_channel(GpioAddress address) {
...
+    .DMA_MemoryDataSize = DMA_MemoryDataSize_HalfWord,
+    .DMA_Mode = DMA_Mode_Circular,
+    .DMA_Priority = DMA_Priority_High,
+    .DMA_M2M = DMA_M2M_Disable,
+  };
+  DMA_Init(DMA1_Channel1, &dma_init);
 
-  // Initialize each adc store with it's own channel for easy initialization
-  for (uint8_t channel = 0; channel < NUM_ADC_CHANNELS; channel++) {
-    s_adc_stores[channel].channel = channel;
```
JarvisWeng, line 177: Perhaps just need to explicitly set all `s_adc_ranks` to 0?

libraries/ms-common/src/arm/adc.c
```diff
@@ -146,156 +117,150 @@ StatusCode adc_add_channel(GpioAddress address) {
...
-
-    // Once conversion is finished, we will receive the semaphore from ISR
-    mutex_lock(&s_adc_status.converting, ADC_TIMEOUT_MS);
-    mutex_unlock(&s_adc_status.converting);
-  }
-
-  *reading = s_adc_stores[channel].reading;
-
+  // Index of reading is its rank - 1
+  *reading = s_adc_readings[s_adc_ranks[channel] - 1];
```
JarvisWeng, line 229: So ranks are 1-indexed?

libraries/ms-common/src/x86/adc.c
```diff
@@ -118,125 +96,104 @@ static StatusCode prv_check_channel_enabled(uint8_t channel) {
...
-  // Initialize all channels registered via adc_add_channel()
-  for (uint8_t index = 0; index < s_adc_status.active_channels; index++) {
-    // Each channel is configured with ascending rank, starting at one
-    uint8_t rank = index + 1;
-  }
-
-  // By default, enable vref and temp sensor for voltage conversions
   adc_add_channel(ADC_REF);
 
+  s_adc_readings[s_adc_ranks[ADC_Channel_TempSensor] - 1] = MOCK_VREFINT;
```
JarvisWeng, line 128: It looks like this is the only place that `s_adc_readings` is set? How are other readings set?

libraries/ms-common/src/arm/adc.c
mitchellostler: I accidentally left two in, but I'm not sure which one is the correct placement. Let's just leave it for now since I know it works, and it won't have negative effects

libraries/ms-common/src/arm/adc.c
mitchellostler: Yup

libraries/ms-common/src/arm/adc.c
mitchellostler: Temperature is the only special case where we convert to a value that's not a mV. Vrefint is handled differently on the F1

libraries/ms-common/src/x86/adc.c
mitchellostler: They're set by the adc_set_reading method

libraries/ms-freertos/inc/FreeRTOSConfig.h
mitchellostler: Just allows help for debugging. You can use it to see where the stack is at, I was using it to debug stack overflows

libraries/ms-common/src/arm/adc.c
mitchellostler: arm-none-gcc does this by default for static variables as far as I know

libraries/ms-common/src/arm/adc.c
JarvisWeng: Hm ok

libraries/ms-common/src/x86/adc.c
JarvisWeng: Oh this is an x86 only function.
Arm is done through DMA right?

JarvisWeng (approved):
Yeah looks good

PR #204, opened 2023-07-26
Add pd output configs

mitchellostler:


libraries/ms-drivers/src/bts_load_switch.c
```diff
@@ -48,42 +48,39 @@ static StatusCode bts_switch_get_pin_enabled(Bts7xxxPin *pin) {
   }
 }
 
-StatusCode bts_switch_select_state(BtsLoadSwitchOutput *loadSwitch) {
-  switch (loadSwitch->select_pin->pin_type) {
+static StatusCode prv_bts_switch_select_state(BtsLoadSwitchOutput *load_switch) {
```
JarvisWeng, line 51: Curious, why does it need to be static

projects/power_distribution_front/src/outputs.c
```diff
@@ -0,0 +1,46 @@
+#include "outputs.h"
```
JarvisWeng, line 1: Are you going to create a `power_distribution_back` as well? If so there's a parameter in the SConstruct that defaults to `FRONT` for PD. You might want to remove that then

libraries/ms-drivers/src/bts_load_switch.c
mitchellostler: It's only used in this file, so I made it part of the internal interface

projects/power_distribution_front/src/outputs.c
mitchellostler: I was planning on just leaving it as front, since for now they're identical, and HW is trying to merge the two boards. is there anything in the build system that requires us to add this to the rear pd?

JarvisWeng: Well specifically it's this location flag
https://github.com/uw-midsun/fwxv/blob/b06cb4b1d5170d6841c053fd3061ea831c426a2b/SConstruct#L67

It's used here to just append back/front so that CAN can autogenerated them
https://github.com/uw-midsun/fwxv/blob/b06cb4b1d5170d6841c053fd3061ea831c426a2b/scons/build.scons#L134

JarvisWeng: I suppose if they're combining them the yamls will need to be combined too

JarvisWeng (approved):

PR #206, opened 2023-07-29
Solar Sense mppt draft

Citrusboa:
Boiler plate project for Solar Sense


libraries/codegen/boards/solar_sense.yaml
```diff
@@ -0,0 +1,115 @@
+# Message Definitions in plaintext are on the wiki at:
```
JarvisWeng, line 1: You'll need to add `solar_sense` into the `boards.yaml` I believe

PR #219, opened 2023-10-11
added cruise control monitor to centre console

harsh-patell:


projects/centre_console/src/cc_monitor.c
```diff
@@ -0,0 +1,77 @@
...
+        ((state.target_velocity * CONVERT_VELCOITY_TO_KPH) + 1) / CONVERT_VELCOITY_TO_KPH;
+  } else if (steering_event == STEERING_CC_DECREASE_SPEED_EVENT) {
+    state.target_velocity =
+        ((state.target_velocity * CONVERT_VELCOITY_TO_KPH) - 1) / CONVERT_VELCOITY_TO_KPH;
+  } else if (steering_event == STEERING_REGEN_BRAKE_EVENT) {
+    state.regen_braking = true;
+  }
+}
+
+void send_message_to_mci() {
```
mitchellostler, line 61: Can we call this update_drive_output?

projects/centre_console/inc/cc_monitor.h
```diff
@@ -0,0 +1,26 @@
...
+
+typedef struct LocalState {
+  uint8_t drive_state;
+  bool cc_toggle;
+  bool regen_braking;
+  uint32_t target_velocity;
+} LocalState;
+
+typedef enum {
+  STEERING_REGEN_BRAKE_EVENT = 0,
```
mitchellostler, line 16: Hey, we've had some updates to steering, so this won't be a enum value anymore. 
We can receive multiple events in the same message, and so need to check each one with a mask. We also won't be receiving regen braking from steering, it will be a button press from centre console. 

This should be updated to the following, and moved to the top of the file
```
 #define  STEERING_CC_TOGGLE_MASK 0x1
 #define   STEERING_CC_INCREASE_SPEED_MASK 0x2
 #define   STEERING_CC_DECREASE_SPEED_MASK 0x4
 ```

You can check if an event has happened with the following:
uint8_t cc_info = get_steering_info_digital_input();
if (cc_info & STEERING_CC_TOGGLE_MASK) { // CC toggle event has occurred
...
}

File: projects/centre_console/inc/cc_monitor.h [hunk 3]
```diff
@@ -0,0 +1,26 @@
...
+  STEERING_CC_TOGGLE_EVENT,
+  STEERING_CC_INCREASE_SPEED_EVENT,
+  STEERING_CC_DECREASE_SPEED_EVENT,
+  NUM_STEERING_EVENTS,
+} SteeringDigitalEvent;
+
+void monitor_cruise_control();
```
mitchellostler, line 23: The only function you should have in the header file is the one that is being called from the main loop

projects/centre_console/src/cc_monitor.c
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void update_state() {
+  if (get_pedal_output_brake_output()) {
+    state.cc_toggle = false;
+  }
+
+  // check steering message for CC event (toggle/increase/decrease) or regen braking
+  SteeringDigitalEvent steering_event = get_steering_info_digital_input();
+
```
mitchellostler, line 33: There is one more check you should do here since the regen brake button is now on centre_console. 
With some changes Ayo is making, when the button is pressed it will send a notification to this task. This is not implemented on main yet, so for now add 
`#define REGEN_BRAKE_EVENT_PRESSED 0`
at the top, and then add the following logic in your update_state() function:

```
// Check for regen brake event
uint32_t master_task_notif;
notify_get(&master_task_notif);
if (notify_check_event(&master_task_notif, REGEN_BRAKE_EVENT_PRESSED)) {
  state.regen_braking = !state.regen_braking;
  if (state.regen_braking) {
    pca9555_gpio_set_state(&regen_brake_light, PCA9555_GPIO_STATE_LOW);
  }
}

mitchellostler (commented):
Overall looks good! the structure that you have is correct. There's a couple of changes I've requested, for your original logic and for some new changes that have been made to centre_console/steering

mitchellostler (approved):

PR #220, opened 2023-10-11
implement max17261 driver

03predo:


libraries/ms-drivers/inc/max17261_fuel_gauge_defs.h
```diff
@@ -0,0 +1,102 @@
+#pragma once
+// reg map is on P.27:
+// https://www.analog.com/media/en/technical-documentation/user-guides/max1726x-modelgauge-m5-ez-user-guide.pdf
+typedef enum {
+  MAX17261_STATUS = 0x00,
+  MAX17261_VOLT_ALRT_THRSH,
```
JarvisWeng, line 6: Do most of these just increment?

libraries/ms-drivers/src/max17261_fuel_gauge.c
```diff
@@ -0,0 +1,80 @@
...
+  status_ok_or_return(i2c_read(storage->settings.i2c_port, storage->settings.i2c_address,
+                               (uint8_t *)value, sizeof(uint16_t)));
+  return STATUS_CODE_OK;
+}
+
+StatusCode max17261_set_reg(Max17261Storage *storage, Max17261Registers reg, uint16_t value) {
+  uint8_t buf[3];
+  buf[0] = reg;
+  // send LSByte then MSByte as per datasheet
+  buf[1] = value & 0x00FF;
```
JarvisWeng, line 22: I feel like there's a function for swapping endianness.
`#include <byteswap.h>`
though that should be fine, just a nit

libraries/ms-drivers/src/max17261_fuel_gauge.c
```diff
@@ -0,0 +1,80 @@
+#include "max17261_fuel_gauge.h"
+
+#define PCT_LSB (1U / 256U)                            // LSBit is 1/256%
```
mitchellostler, line 3: This will evaluate to an integer, so will just be zero. 
If you want to do it as float arithmetic, you need to set one of the values as a float (ie 1.0 / 256)
Same goes for values below

libraries/ms-drivers/src/max17261_fuel_gauge.c
mitchellostler: I wouldn't worry about linking in a whole library to do this, this should be fine

libraries/ms-drivers/inc/max17261_fuel_gauge.h
```diff
@@ -0,0 +1,43 @@
...
+  uint16_t charge_term_current;  // LSB = 1.5625 (micro Volts / R Sense)
+
+  uint16_t r_sense_ohms;
+} Max17261Settings;
+
+typedef struct {
+  Max17261Settings settings;
+} Max17261Storage;
+
+StatusCode max17261_get_reg(Max17261Storage *storage, Max17261Registers reg, uint16_t *value);
```
mitchellostler, line 36: Can we get comments giving more info on what each of these are returning, and with units :) 
I see that you set the var names to have some of this info, but would be great to have it more verbose.

mitchellostler (changes requested):
Overall looks great. Just the one issue with the integer types in the defines. 

I also asked for comments, but if you want to write up some documentation that covers how it's initialized, how normal operation works, and what data we have available, that.would be even better. I've created a page [here] (https://uwmidsun.atlassian.net/l/cp/JLPESZnT) which you can use if you like

libraries/ms-drivers/inc/max17261_fuel_gauge_defs.h
mitchellostler: I'm also confused, are there register values that are skipped in between? or are they just ones you deemed we wouldn't use?

libraries/ms-drivers/src/max17261_fuel_gauge.c
```diff
@@ -0,0 +1,80 @@
...
+#define CAP_LSB (5U / storage->settings.r_sense_ohms)  // LSBit is 5 micro Volt hrs / Rsense
+#define TIM_LSB (5625U)                                // LSBit is 5625ms
+
+StatusCode max17261_get_reg(Max17261Storage *storage, Max17261Registers reg, uint16_t *value) {
+  // unsure of underlying type of enum, cast to uint8_t to be sure
+  uint8_t reg8 = reg;
+  status_ok_or_return(i2c_write(storage->settings.i2c_port, storage->settings.i2c_address,
+                                (uint8_t *)&reg8, sizeof(uint8_t)));
```
mitchellostler, line 11: Nit: These uint8_t* casts aren't necessary but won't hurt. Usually good to not use casts you don't need though

libraries/ms-drivers/inc/max17261_fuel_gauge.h
mitchellostler, line 36: Are we expecting to call these functions directly?
If we're just going to be calling them through the other functions, might as well not include them in the header

libraries/ms-drivers/inc/max17261_fuel_gauge_defs.h
03predo: Yeah so most of them just increment but some of them are reserved. You can see the reg map on P.27 of this datasheet https://www.analog.com/media/en/technical-documentation/user-guides/max1726x-modelgauge-m5-ez-user-guide.pdf

mitchellostler (approved):

PR #224, opened 2023-10-28
Fault task

Bafran:
We were discussing how we wanted to add fault behaviour and I thought this was a good option. This task basically will just take over if its "watchdog" fails. The handler is defined per project and passed as a function pointer. I thought this was needed since BMS/CC/MCI etc. all need to do different things and the main thing that can be made into a module is this task.

If we need more granularity like monitoring a GPIO I could prob add that into this as well but lmk what u think

Still gotta actually add the BMS behaviour but it would essentially just be a list of messages. (Ill delete the smoke test after review)

### DIFF
--- a/libraries/ms-common/src/fault.c
+++ b/libraries/ms-common/src/fault.c
@@ -1,0 +1,54 @@
+#include "fault.h"
+
+#include "delay.h"
+#include "log.h"
+#include "semphr.h"
+#include "status.h"
+#include "tasks.h"
+
+#define FAULT_CYCLES 5
+
+static FaultStorage *s_storage;
+
+static uint8_t s_max_cycles_missed;
+static uint32_t s_cycles_missed_bitset;
+
+TASK(fault, TASK_MIN_STACK_SIZE) {
+  bool fault = false;
+  s_max_cycles_missed = 0;
+
+  while (true) {
+    if (!fault) {
+      // Check faults
+      for (uint8_t i = 0; i < s_storage->num_faults; i++) {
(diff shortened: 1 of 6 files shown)


libraries/ms-common/src/fault.c
```diff
@@ -0,0 +1,54 @@
...
+
+static uint8_t s_max_cycles_missed;
+static uint32_t s_cycles_missed_bitset;
+
+TASK(fault, TASK_MIN_STACK_SIZE) {
+  bool fault = false;
+  s_max_cycles_missed = 0;
+
+  while (true) {
+    if (!fault) {
```
JarvisWeng, line 21: Hm, could you just wait on a `queue`, so you don't busy wait here?

smoke/fault/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+int main() {
+   tasks_init();
+   log_init();
+   LOG_DEBUG("Welcome to TEST!");
+
+   FaultStorage storage = {
+      // Watch critical faults
+      // If any of these are non-zero, the fault task takes over the ECU and
+      // will require a power cycle
+      .faults = {
```
JarvisWeng, line 48: How will `faults` be set?

libraries/ms-common/src/fault.c
ShiCheng-Lu: if (!fault) path is taken, this only runs every once every 100 ms, I think the issue is if there is a fault, this would be constantly calling the handler.

PR #226, opened 2023-10-29
reduce can mem

ShiCheng-Lu:
- ordering of variables in can_rx_struct

```
  28256    1408   13796   43460    a9c4 build/arm/bin/projects/centre_console
```

### DIFF
--- a/libraries/codegen/templates/_rx_structs.h.jinja
+++ b/libraries/codegen/templates/_rx_structs.h.jinja
@@ -11,6 +11,8 @@ typedef struct
     {%- for signal, signal_data in message_data['signals'].items() %}
     uint{{signal_data['length']}}_t {{message}}_{{signal}};
     {%- endfor %}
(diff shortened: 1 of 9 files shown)


libraries/codegen/templates/_rx_structs.h.jinja
```diff
@@ -11,6 +11,8 @@ typedef struct
     {%- for signal, signal_data in message_data['signals'].items() %}
     uint{{signal_data['length']}}_t {{message}}_{{signal}};
     {%- endfor %}
+{%- endfor %}
```
JarvisWeng, line 14: What's this for?

ShiCheng-Lu: variable packing, so putting the "received_{{message}}" variables together at the end saves space.

JarvisWeng (approved):

PR #227, opened 2023-11-01
Logs working

mitchellostler:
Just needed to disable uart_rx

### DIFF
--- a/libraries/ms-common/src/arm/uart.c
+++ b/libraries/ms-common/src/arm/uart.c
@@ -68,6 +68,7 @@ StatusCode uart_init(UartPort uart, UartSettings *settings) {
 
   USART_InitTypeDef usart_init;
   USART_StructInit(&usart_init);
+  usart_init.USART_Mode = USART_Mode_Tx;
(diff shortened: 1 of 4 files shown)


ShiCheng-Lu:
sounds a bit hacky, but If we don't ever need uart_rx I think its fine.

libraries/ms-common/src/arm/uart.c
```diff
@@ -77,10 +78,6 @@ StatusCode uart_init(UartPort uart, UartSettings *settings) {
   // Init with disabled TX interrupts otherwise we will get TX
   // buffer empty continuously when there is no data to send
   USART_ITConfig(s_port[uart].base, USART_IT_TXE, DISABLE);
-  // Init with enabled RX interrupts becuase they are called only
-  // when there is data to be received
-  USART_ITConfig(s_port[uart].base, USART_IT_RXNE, ENABLE);
```
JarvisWeng, line 82: Maybe just comment this out?

JarvisWeng (approved):

PR #232, opened 2023-11-10
Fwxv 999 warrenxu 103 hw

warren-xu:
fw 102 tasks


projects/queues/src/main.c
```diff
@@ -0,0 +1,65 @@
...
+    ret = queue_send(&s_queue1, &to_send, 0);
+    if (ret != STATUS_CODE_OK) {
+      LOG_DEBUG("write to queue failed\n");
+    }
+    to_send++;
+  }
+}
+TASK(task2, TASK_STACK_512) {
+  LOG_DEBUG("Task 2 initialized!\n");
+  const char outstr[ITEM_SZ];
```
JarvisWeng, line 40: You should initialize your array like so
`const char outstr[ITEM_SZ] = {};`

projects/queues/src/main.c
```diff
@@ -0,0 +1,65 @@
...
+  char *read = outstr;
+  StatusCode ret;
+  uint32_t receive = 0;
+  while (true) {
+    // Your code goes here
+    ret = queue_receive(&s_queue1, &receive, 1000);
```
JarvisWeng, line 46: This `&receive`, doesn't really make sense. You're trying to put the item you receive in `receive`, but the receive size is only 4 bytes, while the item size is 6 bytes

projects/queues/src/main.c
```diff
@@ -0,0 +1,62 @@
...
+  .num_items = QUEUE_LEN,
+  .storage_buf = s_queue1_buf,
+};
+TASK(task1, TASK_STACK_512) {
+  LOG_DEBUG("Task 1 initialized!\n");
+  StatusCode ret;
+  uint32_t to_send = 0;
+  while (true) {
+    // Your code goes here
+    ret = queue_send(&s_queue1, s_list[to_send], 0);
```
JarvisWeng, line 28: You want to use `&s_list[to_send]` because this function takes in the `address` of the value you want to send, not the value itself

projects/queues/src/main.c
```diff
@@ -0,0 +1,62 @@
...
+    if (ret != STATUS_CODE_OK) {
+      LOG_DEBUG("write to queue failed\n");
+    }
+    delay_ms(100);
+    ++to_send;
```
JarvisWeng, line 33: What about when `to_send` is greater than 5? What happens then?

projects/queues/src/main.c
```diff
@@ -0,0 +1,68 @@
...
+  StatusCode ret;
+  uint32_t to_send = 0;
+  while (true) {
+    // Your code goes here
+    ret = queue_send(&s_queue1, &s_list[to_send], 0);
+
+    delay_ms(100);
+    ++to_send;
+    if (to_send >= 5) {
+      ret = STATUS_CODE_OUT_OF_RANGE;
```
JarvisWeng, line 33: You should just use `% QUEUE_LEN`. That'll just go back to the first item

PR #233, opened 2023-11-10
Centre Console Buttons

ay0fasan:


libraries/master/inc/master_task.h
```diff
@@ -9,6 +9,7 @@ void set_slow_cycle_count(uint32_t cycles);
 void run_fast_cycle();
 void run_medium_cycle();
 void run_slow_cycle();
+void pre_loop_init();
```
mitchellostler, line 12: @JarvisWeng I added this function to master task

libraries/ms-common/src/arm/i2c.c
```diff
@@ -299,19 +299,20 @@ static void prv_ev_irq_handler(I2CPort i2c) {
...
+        // If we only have one byte left, don't ack slave
+        // This finishes the transaction after last byte sent
+        I2C_AcknowledgeConfig(s_port[i2c].base, DISABLE);
+        I2C_GenerateSTOP(s_port[i2c].base, ENABLE);
+      } else if (s_port[i2c].num_rx_bytes == 0) {
+        // Unlock mutex after receiving last byte
+        I2C_ITConfig(s_port[i2c].base, I2C_IT_ERR | I2C_IT_EVT, DISABLE);
+        I2C_GenerateSTOP(s_port[i2c].base, ENABLE);
+        xSemaphoreGiveFromISR(s_port[i2c].i2c_buf.wait_txn.handle, &xTaskWoken);
+      }
```
mitchellostler, line 315: Extra bytes were being received by rx past what we were supposed to be getting due to the queue_send being outside of the check for number of bytes to send. This was breaking read

libraries/master/inc/master_task.h
JarvisWeng: Sure but I guess why is this necessary? Shouldn't all initializations be done in main?

projects/centre_console/src/cc_buttons.c
```diff
@@ -0,0 +1,33 @@
+#include "cc_buttons.h"
+
+// Notifies drive/power task of button press event
+StatusCode get_button_press(void) {
+  uint8_t pca9555_reg_val = 0;
+  // TODO: Migrate this read to PCA9555
+  i2c_read_reg(I2C_PORT_1, PCA9555_I2C_ADDR, INPUT1, &pca9555_reg_val, 1);
+
+  if (pca9555_reg_val == PCA9555_REG_DEFAULT) {  // No button pressed
```
JarvisWeng, line 9: Why is default the only one that doesn't need to flip `~pca9555_reg_val`

projects/centre_console/src/main.c
```diff
@@ -43,14 +56,17 @@ void run_slow_cycle() {}
...
+    .sda = I2C1_SDA,
+    .scl = I2C1_SCL,
+  };
+  i2c_init(I2C_PORT_1, &i2c_settings);
   can_init(&s_can_storage, &can_settings);
 
   LOG_DEBUG("Welcome to TEST! \n");
   fsm_shared_mem_init();
 
-  init_drive_fsm();
```
JarvisWeng, line 51: What was wrong with initializing here?

libraries/master/inc/master_task.h
mitchellostler: Sorry I should have clarified. For things like pca_init, we need i2c operations to do them initially, which require the scheduler to be started. 
Some initialization needs to happen before scheduler start (in main) and some needs to happen after (in pre_loop_init)

projects/centre_console/src/cc_buttons.c
mitchellostler: This check is unecessary, I will remove it. 

It was originally to check if the returned value was equal to 0xff (since the buttons are active low)
The checks below flips the reg to check which bits have been set to 0

ShiCheng-Lu:
test failing due to I2C not being initialized in test_drive_fsm.c, 
```
static I2CSettings i2c_setting = {
  .speed = I2C_SPEED_FAST,
};

void setup_test(void) {
  log_init();
  i2c_init(I2C_PORT_1, &i2c_setting);
}
```
you can add this bit to fix it

Kyle02 (approved):
Asked Mitchell and he said it looks good.

Kyle02 (approved):
Asked Mitchell and he said it looks good.

PR #235, opened 2023-11-14
lights fsm working

devAdhiraj:


libraries/master/src/master_task.c
```diff
@@ -47,9 +47,9 @@ void check_late_cycle(BaseType_t delay) {
 }
 
 TASK(master_task, TASK_STACK_512) {
-  TickType_t xLastWakeTime = xTaskGetTickCount();
   uint32_t counter = 1;
   pre_loop_init();
+  TickType_t xLastWakeTime = xTaskGetTickCount();
```
JarvisWeng, line 52: Any reason for moving this?

libraries/ms-common/inc/soft_timer.h
```diff
@@ -27,6 +27,13 @@ typedef void (*SoftTimerCallback)(SoftTimerId id);
 // the time has expired.
 StatusCode soft_timer_start(uint32_t duration_ms, SoftTimerCallback callback, SoftTimer *timer);
```
JarvisWeng, line 28: Hm having multiple functions may confuse people. Perhaps get rid of this one then?

projects/power_distribution/src/lights_fsm.c
```diff
@@ -1,44 +1,86 @@
...
-    soft_timer_start(SIGNAL_BLINK_PERIOD_MS, prv_lights_signal_blinker, &s_timer_single);
-  } else {
-    // in the initial state light_id_callback == NUM_EE_LIGHT_TYPES
-    gpio_set_state(&RIGHT_LIGHT_ADDR, GPIO_STATE_LOW);
-    gpio_set_state(&LEFT_LIGHT_ADDR, GPIO_STATE_LOW);
-    return;
+  switch (light_id_callback) {
+    case EE_LIGHT_TYPE_SIGNAL_LEFT:
+      left_signal_state ^= 1;
+#ifdef LIGHTS_FSM_DEBUG
```
JarvisWeng, line 27: Are you keeping all these debugs?

libraries/ms-common/inc/soft_timer.h
devAdhiraj: I didn't want to get rid of the original function because its used in a bunch of places and it seems to be working fine for those cases.

projects/power_distribution/src/lights_fsm.c
devAdhiraj: I wasn't sure if I should, so I added ifdefs around them, but I can just get rid of them to make it cleaner.

libraries/master/src/master_task.c
devAdhiraj: I was getting an `FSM TIMEOUT` until I moved `pre_loop_init()`. I feel like maybe pre_loop_init() takes up a long time (inits all pd outputs?) so maybe it is messing up timing which is why it fixed it. But I'm not sure tbh

JarvisWeng: Lol I guess that's ok. Would be good to know why though

JarvisWeng (approved):

mitchellostler (approved):

PR #238, opened 2023-11-21
Update Output Config spec for PD based on new rev

thiendan:


mitchellostler (commented):
Hey, I think you committed the wrong files :)

thiendan:
Sorry! I fixed it, it has the right files now

mitchellostler (approved):

PR #241, opened 2023-11-25
SPI Validation

mitchellostler:


SConstruct
```diff
@@ -241,6 +241,11 @@ if PLATFORM == 'arm' and TYPE == 'project':
         
     # flash the MCU using openocd
     def flash_run(target, source, env):
+        import serial
```
mitchellostler, line 244: I want to merge this one in, but this check for the serial path will prevent macs from flashing. We can add a 
```
import platform
if platform.system() == 'Darwin':
    # check for /dev/tty.usbmodem* device
else:
   # check for /dev/serial/
```

And that should resolve. Unfortunately I don't have a dongle, so I can't test at the moment

mitchellostler: We could also move this to another PR if you like

libraries/ms-common/src/arm/spi.c
```diff
@@ -107,92 +105,56 @@ StatusCode spi_init(SpiPort spi, const SpiSettings *settings) {
...
       return STATUS_CODE_RESOURCE_EXHAUSTED;
     }
   }
 
-  // Enable interrupts and start transaction
-  // Transaction started when we spi is enabled, and tx register is empty
-  SPI_I2S_ITConfig(s_port[spi].base, SPI_I2S_IT_ERR | SPI_I2S_IT_TXE, ENABLE);
-  SPI_Cmd(s_port[spi].base, ENABLE);
+  // set spi CS state LOW
+  gpio_set_state(&s_port[spi].cs, GPIO_STATE_LOW);
```
mitchellostler, line 130: +  SPI_Cmd(s_port[spi].base, ENABLE);                                                                    
+  // set spi CS state LOW                                                                    
+  gpio_set_state(&s_port[spi].cs, GPIO_STATE_LOW); 

I had to change it to this on my branch so that the spi cmd happens before the pin state gets set to get the correct behaviour for the CS pin

PR #245, opened 2023-12-12
Fwxv 252 current sense module

Bafran:


projects/bms_carrier/src/current_sense.c
```diff
@@ -2,96 +2,140 @@
...
+
+    .charge_term_current = CHARGE_TERMINATION_CURRENT,
+    .design_capacity = MAIN_PACK_DESIGN_CAPACITY,
+    .empty_voltage = MAIN_PACK_EMPTY_VOLTAGE,
+
+    // Expected MAX current / (uV / uOhmsSense) resolution
+    .i_thresh_max =
+        ((CURRENT_SENSE_MAX_CURRENT) / (ALRT_PIN_V_RES_MICRO_V / CURRENT_SENSE_R_SENSE_U_OHMS)),
+    // Expected MIN current / (uV / uOhmsSense) resolution
+    .i_thresh_min =
```
Bafran, line 127: Could you guys check over the datasheet for me? Might be completely missing something but we need min current to be -27A, this register looks like it holds unsigned positive values?

Page 26, Table 15. https://www.analog.com/media/en/technical-documentation/data-sheets/max17261.pdf

projects/bms_carrier/src/current_sense.c
```diff
@@ -2,96 +2,140 @@
...
-  ads1259_get_conversion_data(&s_ads1259_storage);
-  // Kick new soft timer
-  soft_timer_init_and_start(s_current_storage->conv_period_ms, prv_periodic_ads_read, &s_timer);
-
-  // update average
-  int32_t sum = 0;
-  for (uint16_t i = 0; i < NUM_STORED_CURRENT_READINGS; i++) {
-    sum += s_current_storage->readings_ring[i];
+  if (status != STATUS_CODE_OK) {
+    // TODO (Adel): Handle a fuel gauge fault
```
Bafran, line 51: Also not sure if this is needed. If the fuel gauge fails should there be a retry? Should we open relays at all?

projects/bms_carrier/src/current_sense.c
ShiCheng-Lu: image

says the default is 0x7F80, where high bits 7F is the max value, low bits 80 is the min value, so these are indeed signed 8 bit numbers

03predo (approved):

PR #246, opened 2023-12-13
Cleanup scons mess

ShiCheng-Lu:


SConstruct
```diff
@@ -1,150 +1,117 @@
...
-
-AddOption(
-    '--library',
-    dest='library',
-    type='string',
-    action='store',
+    type='choice',
+    choices=("arm", "x86"),
+    default='arm',
+    help="Specifies target platform. One of 'arm' or 'x86'. Defaults to 'arm' if not provided."
```
ShiCheng-Lu, line 26: --platform=... is now checked by scons (optparse) to make sure it's either x86 or arm

SConstruct
```diff
@@ -1,150 +1,117 @@
...
-)
-AddOption(
-    '--py',
-    dest='python',
-    type='string',
-    action='store'
+    '--library', '--project', '--smoke', '--python', '--py',
+    type="string", dest="name",
+    action='callback', callback=set_target,
+    help="Specify the target.   e.g. --library=ms-common, --project=leds, --smoke=adc, --python=example",
```
ShiCheng-Lu, line 33: --library=... / --project=... is now a single option that stores the target in to TARGET.

SConstruct
```diff
@@ -1,150 +1,117 @@
...
+    help="Add CPP defines to a build.   e.g. --define='LOG_LEVEL=LOG_LEVEL_WARN'"
 )
 
 AddOption(
-    '--name',
-    dest='name',
+    '--task',
+    dest='task',
     type='string',
     action='store'
```
ShiCheng-Lu, line 56: changed the option for create new task to `scons new --project=... --task=task_name`

SConstruct
```diff
@@ -1,150 +1,117 @@
...
+    help="(arm) Reports the memory space after a build."
+)
+
+AddOption(
+    '--sanitizer',
+    dest='sanitizer',
+    type='choice',
+    choices=("asan", "tsan"),
+    default="none",
+    help="(x86) Specifies the sanitizer. One of 'asan' for Address sanitizer or 'tsan' for Thread sanitizer. Defaults to none."
```
ShiCheng-Lu, line 72: sanitizer argument to a choice

smoke/adc/src/main.c
```diff
@@ -1,55 +1,50 @@
 #include <stdio.h>
```
ShiCheng-Lu, line 1: file formatted, no changes

SConstruct
```diff
@@ -1,150 +1,117 @@
...
 )
 
 PLATFORM = GetOption('platform')
-PROJECT = GetOption('project')
-LIBRARY = GetOption('library')
-PYTHON = GetOption('python')
-MEM_REPORT = GetOption('mem-report')
-NAME = GetOption('name')
+TARGET = GetOption('name')
```
Kyle02, line 76: Does this get the output from set_target?

SConstruct
```diff
@@ -153,114 +120,77 @@ BUILD_DIR = Dir('#/build').Dir(PLATFORM)
...
 ###########################################################
 # Testing
 ###########################################################
-SConscript('scons/test.scons', exports='VARS')
+if COMMAND == "test":
+    # Add flags when compiling a test
+    TEST_CFLAGS = ['-DMS_TEST=1']
+    env['CCFLAGS'] += TEST_CFLAGS
+    SConscript('scons/test.scons', exports='VARS')
+    SConscript('scons/build.scons', exports='VARS')
```
Kyle02, line 133: Are we deleting the build section and moving into this testing area? Do we not need SConscript('scons/build.scons', exports='VARS') for stuff without "test"

SConstruct
```diff
@@ -153,114 +120,77 @@ BUILD_DIR = Dir('#/build').Dir(PLATFORM)
...
+if PLATFORM == 'arm' and TARGET:
+    project_bin = BIN_DIR.File(TARGET + '.bin')
     # display memory info for the project
-    if MEM_REPORT:
-        get_mem_report = Action("python3 scons/mem_report.py " + "build/arm/bin/projects/{}".format(TARGET))
-        env.AddPostAction(proj_bin(TARGET, False), get_mem_report)
-        
+    if GetOption('mem-report'):
+        AlwaysBuild(Command("#/mem-report", project_bin,
+                            f"python3 scons/mem_report.py build/arm/bin/{TARGET}"))
```
Kyle02, line 186: does TARGET include the /projects/ that was there previously? Or does it not need it anymore

build_system.md
```diff
@@ -10,48 +10,104 @@ Each project and library may also contain a `config.json` file with certain fiel
...
+    lint
+        Lint a target, or all targets if not specified. uses pylint for python and cpplint for c.
+        - e.g. `scons lint`
+        - e.g. `scons lint <target>`
+
+    clean
+        Delete the `build` directory.
+
+Targets:
+targests can be specified with an option.
```
Kyle02, line 82: NIT: "targets"

Kyle02: Also maybe add a full example for sim and test, since itll easier to just copy paste and fill in the correct project. I think itll be easier for people to see whats going on if theres an example like the previous build_system.md 

e.g `scons sim --project=leds`

libraries/codegen/boards/centre_console.yaml
```diff
@@ -11,7 +11,7 @@
 ---
   Messages:
     cc_power_control:
-      id: 8
+      id: 1
```
Kyle02, line 14: What are the ID changes for?

Kyle02 (commented):
Pausing review at libraries/codegen/generator.py

SConstruct
ShiCheng-Lu: yes, destination of '--library', '--project', '--smoke', '--python', '--py', option is "name" and set via the callback  set_target(option, opt, value, parser):, set on line 14

SConstruct
ShiCheng-Lu: test.scons now sets up emitters (basically adds a callback whenever a target is defined) this emitter will be called when build.scons defines a target (project or lib, etc.) then look into the directory of the target and add any tests that exists.

SConstruct
ShiCheng-Lu: target is now the path from root of fwxv, then within the variant dir, it becomes the path relative to build/platform/bin/

.gitignore
```diff
@@ -23,15 +23,6 @@ __pycache__
...
-*_getters.h
-*_pack_msg.h
-*_rx_all.c
-*_rx_structs.h
-*_setters.h
-*_tx_all.c
-*_tx_structs.h
-*_unpack_msg.h
 *can_board_ids.h
-*can_codegen.h
```
ShiCheng-Lu, line 35: these all reside in build/ folder, already ignored

build_system.md
ShiCheng-Lu: done

libraries/codegen/boards/centre_console.yaml
ShiCheng-Lu: felt like it

.gitignore
JarvisWeng: Yeah they weren't before while I was testing. It's fine to get rid of them

libraries/ms-drivers/inc/ltc6811.h
```diff
@@ -74,17 +74,17 @@ static_assert(sizeof(LtcAfeConfigRegisterData) == 6, "LtcAfeConfigRegisterData m
 
 // COMM Register, refer to LTC6803 datasheet page 31, Table 15
 typedef struct {
-  uint8_t icom0 : 4;
-  uint8_t d0 : 8;
-  uint8_t fcom0 : 4;
+  uint16_t icom0 : 4;
```
JarvisWeng, line 77: This shouldn't matter right? Since we're putting them to be 4 bits right?

JarvisWeng (approved):

libraries/ms-drivers/inc/ltc6811.h
ShiCheng-Lu: I believe the compiler was complaining about the uint8_t  d0 : 8; packing 8 bits with 8 bits. since we are only using 4 bits, it doesn't matter as the packed output is the same

Kyle02 (changes requested):
Have you tried running scons sim yet? I'm trying to run this on a test branch but I'm not able to get scons leds to work.

Kyle02:
> Have you tried running scons sim yet? I'm trying to run this on a test branch but I'm not able to get scons leds to work.

We discovered there is an issue with this. I don't think we should merge this until all scons commands have been verified, so shi cheng will be responsible for testing scons flash and other commands

PR #247, opened 2023-12-13
Better autogen

ShiCheng-Lu:
change the autogen to be called only once to generate all the can files for one project

`autogenner -f {source_dir} -t {all source templates...} -f {header dir} -t {all header templates}`

changed test_runner auto-generator to use jinja

added a python template to generate some cansend functions for can (just preliminary)


libraries/codegen/boards/uv_cutoff.yaml
```diff
@@ -10,7 +10,7 @@
 
 ---
   Messages:
-    uv_cutoff_notification:
+    uv_cutoff_notification1:
```
ShiCheng-Lu, line 13: there are two uv_cutoffs, from power-distribution and uw_cutoff, both to telemetry, and autogen can't handle this, so slight name change. Previously, one of them is just ignored, but this is in fact a different message with different id.

libraries/codegen/generator.py
```diff
@@ -1,242 +1,119 @@
...
+
+            messages.append({
+                "id": message["id"],
+                "name": message_name,
+                "signals": signals,
+                "sender": sender,
+                "receiver": message["target"],
+            })
+
+    return {"Boards": boards, "Messages": messages}
```
ShiCheng-Lu, line 90: autogen will now read all board yamls once, all  necessary message info is stored in this dictionary (list of boards to generate board id, and list of messages which is then filtered by the board at jinja generation)

libraries/codegen/generator.py
```diff
@@ -1,242 +1,119 @@
...
+
+
 def main():
-    parser = OptionParser()
-    # TODO: Get rid of -y option or -b option, redundant
```
ShiCheng-Lu, line 190: got rid of -y option, can't get rid of -b option unless we want to infer the board from the output path, which is a bit ugly. Maybe later with better_better_autogen

libraries/codegen/templates/_filters.h.jinja
```diff
@@ -1,16 +1,16 @@
+{% set board = data["Board"] -%}
+{% set messages = data["Messages"] | selectattr("receiver", "contains", board) | list -%}
```
ShiCheng-Lu, line 2: this bit filters the messages to contain only the relevant ones for the specific board

libraries/test_gen/config.json
```diff
@@ -2,5 +2,6 @@
     "libs": [
         "FreeRTOS",
         "ms-common"
-    ]
+    ],
+    "no_lint": true
```
ShiCheng-Lu, line 6: there are some python lint issues in this library, it's annoying to fix, make a ticket for this.

ShiCheng-Lu:
eventually, I want to have all autogeneration stuff in one place, for can autogen, test auto gen, and new project autogen.

better_better_autogen:

something like
-autogen
    - can
        - src
            - {{board}}_tx.c.jinja
            - {{board}}_rx.c.jinja
        - inc
            - {{board}}_struct.h.jinja
            - ...
    - test
        - test_runner.c.jinja
    - new_project
        - config.json
        - src
            - main.c
        - inc
      ....

then specifying `autogen can -o <folder>` will generate all the templates in the can folder into the output folder. each folder will possibly have some function that retrieves the data needed to generate it.

libraries/codegen/templates/can.py.jinja
```diff
@@ -0,0 +1,44 @@
+{% set boards = data["Boards"] -%}
+{% set messages = data["Messages"] -%}
+
+import struct
```
JarvisWeng, line 4: What is this file for?

py/can.py
```diff
@@ -0,0 +1,219 @@
+import struct
```
JarvisWeng, line 1: Isn't this an autogenerated file?

ShiCheng-Lu: right, I'll add this to gitignore

libraries/codegen/templates/can.py.jinja
ShiCheng-Lu: some functions/helpers to create messages in python to send through can. I should separate it to a different PR

mitchellostler (approved):

PR #249, opened 2023-12-19
PD power seq implementation

devAdhiraj:


libraries/codegen/templates/_rx_all.c.jinja
```diff
@@ -53,11 +54,21 @@ StatusCode check_can_watchdogs() {
...
     }
     {%- endif %}
 {%- endfor %}
-    return STATUS_CODE_OK;
+    return status;
 }
+
+{%- for message, message_data in messages.items() %}
+    {% if message_data["target"][board]["watchdog"] != 0 %}
+uint8_t check_{{message}}_msg_watchdog() {
```
ShiCheng-Lu, line 70: you can probably define these as macros like the can rx_getters and tx_setters since they only return the value

projects/power_distribution/src/power_seq_fsm.c
```diff
@@ -1,59 +1,176 @@
...
+}
 
-static void prv_driver_controls_output(void *context) {
-  LOG_DEBUG("Transitioned to DRIVER_CONTROLS\n");
+static void prv_close_relays_state_input(Fsm *fsm, void *context) {
+  check_pd_fault(fsm);
+  uint8_t bms_relay_state = get_battery_relay_info_state();
+  if (bms_relay_state == EE_RELAY_STATE_CLOSE) {
+    GpioState dcdc_state;
+    if (gpio_get_state(&dcdc_valid_pin, &dcdc_state) == STATUS_CODE_OK &&
```
ShiCheng-Lu, line 54: if gpio_get_state doesn't return ok, this would transition to POWER_STATE_ON?

libraries/codegen/boards/motor_controller.yaml
```diff
@@ -99,6 +99,10 @@
       target:
         centre_console:
           watchdog: 0
+        power_distribution:
+          watchdog: 0
       signals:
         notification:
           length: 8
+        relay_status:
+          length: 8
```
ShiCheng-Lu, line 108: please use mc_status: - precharge_status for the precharge status

projects/power_distribution/src/power_seq_fsm.c
```diff
@@ -1,59 +1,176 @@
...
+    power_context.target_state = POWER_STATE_ON;
+    fsm_transition(fsm, TRANSMIT_BMS_CLOSE_RELAYS);
+  }
 }
 
-static void prv_driver_controls_input(Fsm *fsm, void *context) {}
+static void prv_close_relays_state_output(void *context) {
+  LOG_DEBUG("Transitioned to close relays STATE\n");
+  pd_set_output_group(OUTPUT_GROUP_BMS_RELAYS, OUTPUT_STATE_ON);
+  power_context.latest_state = TRANSMIT_BMS_CLOSE_RELAYS;
```
mitchellostler, line 45: We should be setting a message here to tell BMS to close the relays right?

mitchellostler (approved):

PR #251, opened 2023-12-21
update dashboard validation

ay0fasan:


ShiCheng-Lu:
I personally think unless the three 7-seg display is one module (not made up of three individual displays operable on their own) then they should be split up into individual displays, keep the original structure and function for single 7-seg displays, but add additional functions to drive the displays as a group.

libraries/ms-drivers/src/seg_display.c
```diff
@@ -5,87 +5,132 @@
...
-void set_seg_display(SegDisplay *display, uint16_t val);
-void seg_display_reset(SegDisplay *display);
-void set_digit(SegDisplay *display, GpioAddress *digit, uint16_t val);
-
-StatusCode seg_display_init(SegDisplay *display) {
-  gpio_init_pin(&display->A, GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_LOW);
-  gpio_init_pin(&display->B, GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_LOW);
-  gpio_init_pin(&display->C, GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_LOW);
-  gpio_init_pin(&display->D, GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_HIGH);
+void set_seg_displays(SegDisplay *display, uint16_t val1, uint16_t val2, uint16_t val3);
```
ShiCheng-Lu, line 8: private functions should start with prv_, and define them at the top here instead of a declaration then definition

libraries/ms-drivers/inc/seg_display.h
```diff
@@ -1,26 +1,34 @@
...
   const GpioAddress D2;
+  const GpioAddress A3;
+  const GpioAddress B3;
+  const GpioAddress C3;
   const GpioAddress D3;
+  const GpioAddress DP;
+  const GpioAddress Digit1;
+  const GpioAddress Digit2;
+  const GpioAddress Digit3;
 } SegDisplay;
```
ShiCheng-Lu, line 25: how does this work? is it one pin for each segment or handled by another driver or something. If it's one pin per segment, shouldn't it be something like this?
1200px-7_segment_display_labeled svg

what are the Digit1 Digit2 Digit3 pins? where are EFG

mitchellostler:
> I personally think unless the three 7-seg display is one module (not made up of three individual displays operable on their own) then they should be split up into individual displays, keep the original structure and function for single 7-seg displays, but add additional functions to drive the displays as a group.

@ShiCheng-Lu we needed to make this into one module, as the three 7-seg displays share the digit1-digit3 update pins. 

Each 7-seg display is controlled via an IC driver, which converts 4 gpio outputs (A-D) controlled by the stm32 and representing a binary value, into the hex representation on the display. Each display has its own 4 value pins, which allow us to update the displays independently. 

We can only do this one digit at a time, which the digit1-digit3 pins select.  These pins are shared between the displays to conserve pins. What this results in is that for each of the 3 digits, we need to 

1. Set the value pins for each 7-seg so that the 4 input pins for each represent the value we want to display
2. Toggle the Digit pin to update all of the digits at that index at once (ie the first digit of each display updates at the same time) 
3. Repeat continuously for all digits

projects/centre_console/src/update_dashboard.c
```diff
@@ -8,22 +8,20 @@
 #include "seg_display.h"
 
 // Multiplication factor to convert CAN motor velocity (cm/s) into drive output velocity (mm/s)
```
mitchellostler, line 10: Can you change this to be cm/s to km/hr
our conversion should be (left + right) / (2 * CONVERT_VELOCITY_KM_H)
where convert velocity is 27.78 
(1/100000 km/cm * 3600 sec/hr)

mitchellostler (approved):

PR #252, opened 2024-01-08
Add pd_fault check + indicator on centre console

Akashem06:


projects/centre_console/src/update_dashboard.c
```diff
@@ -90,6 +95,11 @@ void update_indicators(uint32_t notif) {
     }
     s_last_lights_state = get_steering_info_input_lights();
   }
+
+  // Update Aux warning LED
+  if (get_pd_status_fault_bitset() != 0) {
+    pca9555_gpio_set_state(&s_output_leds[AUX_WARNING_LED], PCA9555_GPIO_STATE_HIGH);
+  }
```
mitchellostler, line 102: Need to add an else condition to set it low, otherwise it will never turn off

mitchellostler (changes requested):
Overall looks good, just the one fix

mitchellostler (approved):

PR #253, opened 2024-01-15
Python CAN Parsing 

Akashem06:


mitchellostler (approved):
This all looks good to me. I think I already provided feedback in our meeting so you can go ahead and merge this if it's all good to go. 

As your testing things, and you think of any improvements you want to make to the tool, feel free to add them on! 

I'm excited about this, good work

can/tools/can.py
```diff
@@ -0,0 +1,361 @@
+import struct
+import subprocess
+import time 
+import threading
+
+SYSTEM_CAN_MESSAGE_BABYDRIVER_BABYDRIVER = 2016
```
JarvisWeng, line 6: Hm you could probably autogen a `.py` file and then just import that instead

libraries/codegen/templates/can.py.jinja
```diff
@@ -0,0 +1,57 @@
+{% set boards = data["Boards"] -%}
+{% set messages = data["Messages"] -%}
+
+import struct
+import subprocess
+import time 
+import threading
+
+{% for message in messages -%}
```
JarvisWeng, line 9: Oh if `can.py` is already an autogenerated file then don't commit it

JarvisWeng (approved):

can/tools/can.py
Bafran: If this file is autogenerated, add it to the .gitignore and delete this file so it's not in the PR

PR #254, opened 2024-01-18
PD FSM Fixes

Akashem06:


libraries/ms-common/inc/fsm.h
```diff
@@ -110,7 +110,7 @@ typedef struct Fsm {
   Fsm *name##_fsm = &((Fsm){        \
       .num_states = num_fsm_states, \
   });                               \
-  TASK(name, TASK_STACK_512) {      \
+  TASK(name, TASK_STACK_256) {      \
```
Bafran, line 113: Were there memory issues with this at 512?

mitchellostler: Yeah, I believe adhiraj was troubleshooting this. We should make this configurable.

@Akashem06 Can you add a stack size param to the FSM() macro which we will pass to the TASK() macro. I think that reducing the memory size for this may have affects on all the other project FSMs, and it really shouldn't be hardcoded, that doesn't make sense

projects/fsm_demo/src/main.c
```diff
@@ -25,7 +25,6 @@ void run_medium_cycle() {
   wait_tasks(1);
   fsm_run_cycle(fsm2);
   wait_tasks(1);
-  delay_ms(1000);
```
mitchellostler, line 28: Lets leave this in, unless there's a reason you took it out

projects/power_distribution/src/main.c
```diff
@@ -35,6 +35,8 @@ void pre_loop_init() {
   pd_sense_init();
   adc_init();
   pd_set_active_output_group(OUTPUT_GROUP_POWER_OFF);
+  i2c_init(I2C_PORT_1, &i2c_settings);
```
mitchellostler, line 38: Yeah, I should have caught this on the initial review, my bad

mitchellostler (changes requested):
Just a couple changes, only one functional change. Looks good!

projects/fsm_demo/src/main.c
Akashem06: I got rid of it because it causes Master task to timeout (any delay > 500ms) It seems like the delay in the medium cycle causes mastertasks xTaskDelayUntil() to return PDFalse. 

I'm not 100% sure why but I think it is because the delay_ms causes the scheduled xTaskDelayUntil() to be in the past, returning PDFault. 

If you could explain it better that would be great!

Bafran: Yeah I think removing this line is correct since `xDelayUntil` takes care of the cycle time

Bafran:
Closing this because you said these changes were in your newer PR. If you need to reopen this we can do that

PR #256, opened 2024-01-21
Added simulation project

cnscwong:


projects/simulation/inc/operation_listener.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <arpa/inet.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <linux/can.h>
```
JarvisWeng, line 6: Shouldn't need CAN

projects/simulation/inc/operation_listener.h
```diff
@@ -0,0 +1,38 @@
...
+
+typedef enum {
+  GPIO_SET = 0,
+  GPIO_TOGGLE,
+  GPIO_IT_TRIGGER,
+  ADC_SET_READING,
+  I2C_SET_READING,
+  SPI_SET_RX,
+  UART
+} Operations;
```
JarvisWeng, line 38: Maybe just add a `NUM_OF_OPERATIONS` at the end

projects/simulation/src/main.c
```diff
@@ -0,0 +1,188 @@
...
+    sscanf(buffer, "%d: %d, %[^\n\t]", &operation, &length, buffer);
+    if (operation < 0 || operation > 6) {
+      LOG_DEBUG("Invalid operation: %d\n", operation);
+      continue;
+    }
+
+    input = operation;
+    switch (input) {
+      case GPIO_SET:
+        LOG_DEBUG("GPIO_SET\n");
```
JarvisWeng, line 45: Hm could you just put each operation in a function? It just makes it easier to read

projects/simulation/src/main.c
```diff
@@ -0,0 +1,188 @@
...
+    LOG_DEBUG("Accept failed: %d \n", newsockfd);
+    return 0;
+  }
+  LOG_DEBUG("Accept successful\n");
+
+  sim_init(newsockfd);
+  close(newsockfd);
+  close(socketfd);
+}
+#else
```
JarvisWeng, line 184: There is no if. This should only ever run on x86

projects/simulation/src/main.c
```diff
@@ -0,0 +1,188 @@
...
+        break;
+      default:
+        LOG_DEBUG("UNRECOGNIZED OPERATION: %d\n", input);
+    }
+    usleep(1000000);
+  }
+}
+
+#ifdef x86
+int main(int argc, char *argv[]) {
```
JarvisWeng, line 150: You should be getting the port from the args, but this is fine for now

projects/simulation/src/simulation_test.c
```diff
@@ -0,0 +1,60 @@
...
+#include <unistd.h>
+
+#include "gpio.h"
+#include "gpio_it.h"
+#include "log.h"
+
+#define PORT 2520
+
+int main() {
+  int socketfd;
```
JarvisWeng, line 24: Also consider putting this socket logic into a function so other users don't need to copy this

JarvisWeng (commented):
Overall, pretty good job! Just a couple nits but otherwise its basically what I wanted

PR #257, opened 2024-01-23
Fwxv 255 moved I2C to PCA

warren-xu:


warren-xu:
accidentally included gpio_leds folder in the pull request, can ignore

libraries/ms-drivers/src/pca9555_gpio_expander.c
```diff
@@ -135,3 +135,23 @@ StatusCode pca9555_gpio_subscribe_interrupts(const GpioAddress *interrupt_pin, E
   gpio_it_register_interrupt(interrupt_pin, &interrupt_settings, event, task);
   return STATUS_CODE_OK;
 }
+
+StatusCode pca9555_get_register_state(I2CAddress addr, uint16_t *reg_val) {
+  if (addr >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid PCA9555 port.");
+  }
+  uint8_t reg_arr[2] = {};       // store read value
+  uint8_t reg_to_read = INPUT0;  // read the first register
```
JarvisWeng, line 144: What is `INPUT0` and `INPUT1`? Can't seem to find them.
But basically you'll always only read and write 1 byte?

projects/gpio_leds/README.md
```diff
@@ -0,0 +1,16 @@
+<!--
```
JarvisWeng, line 1: You should probably get rid of this folder

libraries/ms-drivers/src/pca9555_gpio_expander.c
warren-xu: I was trying to find at what address the PCA9555 registers are at, but I couldn't find them so I assumed it was INPUT0 and INPUT1. Do you know where I should be reading from? 
Yeah, essentially the function will read and write one byte for both registers and output them as a 16-bit int.

Bafran: If you look at the datasheet pg 19, it should be there

libraries/ms-drivers/src/pca9555_gpio_expander.c
```diff
@@ -135,3 +135,23 @@ StatusCode pca9555_gpio_subscribe_interrupts(const GpioAddress *interrupt_pin, E
   gpio_it_register_interrupt(interrupt_pin, &interrupt_settings, event, task);
   return STATUS_CODE_OK;
 }
+
+StatusCode pca9555_get_register_state(I2CAddress addr, uint16_t *reg_val) {
+  if (addr >= NUM_I2C_PORTS) {
+    return status_msg(STATUS_CODE_INVALID_ARGS, "Invalid PCA9555 port.");
+  }
+  uint8_t reg_arr[2] = {};     // store read value
+  uint8_t reg_to_read = 0x00;  // read the first register
```
JarvisWeng, line 144: The regs to read are 0x0 and 0x1? Could you make them into #defines so they're not magic numbers?

Bafran:
Closing out for now, please let us know if it should get merged

PR #258, opened 2024-01-26
Pd rev2 fixes

Akashem06:


projects/power_distribution/src/lights_fsm.c
```diff
@@ -4,6 +4,12 @@
 #include "outputs.h"
 #include "power_distribution_getters.h"
 
+#define brake_lights()                                        \
```
mitchellostler, line 7: Not a huge fan of this syntax. I think it would be better to just make it a short prv_ function

mitchellostler (changes requested):
Just the one change. I saw you also have your Power FSM changes up, but it looks like they're in this PR too. If you address the comments on that PR here, we can merge this one and close the other, or we can merge that one first and then you can resolve conflicts, up to you.

projects/power_distribution/src/power_seq_fsm.c
```diff
@@ -72,14 +72,6 @@ static void prv_on_state_output(void *context) {
 
 static void prv_on_state_input(Fsm *fsm, void *context) {
   pd_fault_ok_or_transition(fsm);
-  if (power_context.target_state == POWER_STATE_OFF) {
```
Bafran, line 75: why did this logic get removed?

mitchellostler (approved):

PR #259, opened 2024-02-01
Hw timer

Akashem06:


libraries/ms-common/src/arm/hw_timer.c
```diff
@@ -0,0 +1,46 @@
...
+    .TIM_Prescaler = (clocks.PCLK1_Frequency / 1000000) - 1,  // 1 Mhz
+    .TIM_CounterMode = TIM_CounterMode_Up,
+    .TIM_Period = UINT16_MAX,
+    .TIM_ClockDivision = TIM_CKD_DIV1,
+  };
+  TIM_TimeBaseInit(TIM2, &timer_init);
+  TIM_Cmd(TIM2, ENABLE);
+}
+
+void hw_timer_delay_us(uint32_t us) {
```
Bafran, line 28: So, a possible issue with this is if a task delays for say 2ms, then gets pre empted by a task that delays for 1ms, the higher priority task would reset the timer, complete its delay, and then when the original task takes over, the timer would be disabled. I'm not 100% sure what the best fix for this is but I'll have to think about that

smoke/hw_timer/src/main.c
```diff
@@ -0,0 +1,44 @@
...
+void pre_loop_init() {}
+
+TASK(hw_timer_task, TASK_STACK_512) {
+   gpio_init_pin(&leds[0], GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_LOW);
+   gpio_init_pin(&leds[3], GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_HIGH);
+  while (true) {
+    gpio_toggle_state(&leds[0]);
+    gpio_toggle_state(&leds[3]);
+    LOG_DEBUG("HW Delay");
+    hw_timer_delay_ms(100);
```
Bafran, line 26: Kind of a nitpick, I'd say test a shorter delay like 300us. It better reflects the use case of a module like this. I think you said you'd previously tested that though

Akashem06: I tested with 10us and it worked, but I can change the smoke tests delay

libraries/ms-common/src/arm/hw_timer.c
JarvisWeng: Would it? TIM_GetCounter should be an atomic call. So the unless it happens before `start_count`, it wouldn't be an issue right?
If that's the case, we could critical section this part until later

libraries/ms-common/src/arm/hw_timer.c
```diff
@@ -0,0 +1,54 @@
...
+    uint32_t current_count = TIM_GetCounter(TIM2);
+    if (current_count <= start_count) {
+      elapsed_time += (UINT16_MAX + 1 - start_count) + current_count;
+    } else {
+      elapsed_time += current_count - start_count;
+    }
+    start_count = current_count;
+  }
+
+  TIM_Cmd(TIM2, DISABLE);
```
Bafran, line 47: Add a function for init and enable the timer

Bafran (changes requested):
almost lgtm

Bafran:
Does this still need the change for timer compare?

http://stm32.kosyak.info/doc/group___t_i_m___exported___functions.html#gab3a6031f187cb8af62eb09a67b4fd2ad

PR #260, opened 2024-02-03
Better better autogen

ShiCheng-Lu:
Um, so this needs scons 4.x, which is installed via pip instead of apt. scons 3.6 uses outdated version of python (or maybe python2 instead of python3)
`pip install scons`

Test plan:

none platform specific:
- [X] format + lint
- [x] new project/smoke/task/lib/py

x86:
- [x] build
- [x] test + clean test
- [x] rebuild when project change
- [x] re-autogen when autogen or config change
- [x] sim + clean sim
- [x] build python runs python
- [x] gdb + clean gdb

arm:
- [x] build
- [ ] test + clean test
- [x] rebuild when project change
- [x] re-autogen when autogen or config change
- [ ] flash + clean flash
- [x] mem-report


projects/leds/test/test_led_1.c
```diff
@@ -36,7 +36,11 @@ void test_leds(void) {
     LOG_DEBUG("blink\n");
     for (uint8_t i = 0; i < SIZEOF_ARRAY(leds); i++) {
       gpio_toggle_state(&leds[i]);
+#ifdef x86
+      delay_ms(10);
+#else
       delay_ms(200);
+#endif
```
ShiCheng-Lu, line 43: tests runs too slow on x86 lmao,

PR #264, opened 2024-02-25
Pedal validate

A3sanjay:


projects/pedal/src/main.c
```diff
@@ -26,8 +34,9 @@ const CanSettings can_settings = {
   .loopback = true,
```
JarvisWeng, line 34: Change this to `false` else it won't work on the board

projects/pedal/src/pedal_data.c
```diff
@@ -4,55 +4,59 @@
...
-  status_ok_or_return(max11600_read_raw(s_max11600_storage));
-
-  int32_t range = s_calib_blob->brake_calib.upper_value - s_calib_blob->brake_calib.lower_value;
-
-  int32_t reading_upscaled =
-      (int32_t)s_max11600_storage->channel_readings[channel] * EE_PEDAL_VALUE_DENOMINATOR;
-  reading_upscaled -= s_calib_blob->brake_calib.lower_value * EE_PEDAL_VALUE_DENOMINATOR;
-  reading_upscaled *= 100;
+StatusCode read_pedal_data(int32_t *reading, MAX11600Channel channel) {
+  volatile uint16_t adc_reading;
```
Bafran, line 30: I don't think this needs to be volatile, you're storing a value in this variable from a buffer

scons/common.py
```diff
@@ -27,8 +28,12 @@ def parse_config(entry):
 def flash_run(entry):
     '''flash and run file, return a pyserial object which monitors the device serial output'''
     try:
-        output = subprocess.check_output(["ls", "/dev/serial/by-id/"])
-        device_path = f"/dev/serial/by-id/{str(output, 'ASCII').strip()}"
+        if platform == 'darwin':
```
Bafran, line 31: What were these changes for?

smoke/pedal_calib/src/main.c
```diff
@@ -6,70 +7,50 @@
 #include "i2c.h"
 #include "interrupt.h"
 #include "log.h"
-#include "max11600.h"
 #include "pedal_calib.h"
+#include "pedal_data.h"
 #include "test_helpers.h"
 #include "unity.h"
 
 #define THROTTLE_CHANNEL MAX11600_CHANNEL_0
```
Bafran, line 15: Are these still needed?

smoke/pedal_calib/src/main.c
```diff
@@ -6,70 +7,50 @@
...
 }
 
-int main(void) {
-  log_init();
-  gpio_init();
-  interrupt_init();
-  gpio_it_init();
-  flash_init();
+TASK(smoke_pedal_calib_task, TASK_STACK_512) {
+  gpio_init_pin(&brake, GPIO_ANALOG, GPIO_STATE_LOW);
```
Bafran, line 47: The brake shouldn't need to be calibrated anymore since it's a limit switch

smoke/pedal_calib/src/main.c
```diff
@@ -6,70 +7,50 @@
...
+  gpio_init_pin(&throttle, GPIO_ANALOG, GPIO_STATE_LOW);
+  adc_add_channel(brake);
```
Bafran, line 49: Remove brake

smoke/pedal_calib/src/main.c
```diff
@@ -82,8 +63,22 @@ int main(void) {
   pedal_calib_init(&s_brake_calibration_storage);
```
Bafran, line 63: This should be changed to throttle and brake removed. It doesn't really matter since its just a memset 0 but should be done

Bafran (commented):
Looks good thank you! I know this is working and most of these comments are small code improvements but a few of them definitely need to be addressed.

Bafran (approved):
lgtm

PR #273, opened 2024-04-01
Begin Centre Console Buzzer Project

ejramas:


Akashem06 (changes requested):
minor changes!

It is good practice to make srue your code passes the github workflows!
This means it a) Builds using scons --project=<enter project> b) is formatted using scons format c) has no lint errors using scons lint

PR #276, opened 2024-04-22
Idk why this was set like this

mitchellostler:


JarvisWeng:
Oh right. I think this change does nothing. There's somewhere down the line that we pull that actual clock rate from our SConstruct. I was trying to do this before but just changing these `freeRTOS` files will do nothing

PR #278, opened 2024-05-14
CLI for Controller Board via UART Rx

et312:


projects/uart_cli/src/cli.c
```diff
@@ -0,0 +1,111 @@
...
+        return NULL;
+      }
+      --idx;
+      cmd_buffer[idx % MAX_CMD_LEN] = 0;
+      printf("\b \b");
+    } else {
+      cmd_buffer[idx % MAX_CMD_LEN] = data;
+      ++idx;
+      printf("%c", data);
+      if (idx == MAX_CMD_LEN) {
```
mitchellostler, line 73: If this occurs, you should probably just stop accepting new characters instead of printing a warning. You can add this to the top, where you check against idx, and if they enter more characters, just don't echo anything or put them in the buffer, unless you get a newline.

projects/uart_cli/src/cli.c
```diff
@@ -0,0 +1,111 @@
...
+        printf("WARNING: Character limit reached\n");
+      }
+    }
+  }
+}
+
+void cmd_parse(char *cmd) {
+  char peripheral[MAX_CMD_LEN + 1] = { 0 };
+  strip_ws(cmd);
```
mitchellostler, line 82: I would just add the strip_ws call to inside tok_cmd, since you'll need it for all instances of calling tok_cmd

projects/uart_cli/src/cli.c
```diff
@@ -0,0 +1,111 @@
...
+  tok_cmd(cmd, peripheral);
+
+  if (strcmp(peripheral, "help") == 0 || strcmp(peripheral, "h") == 0 ||
```
mitchellostler, line 85: Should just make the cmd automatically lower case here. This will cover any mixed-case or upper case scenarios, and will mean that you only need to add one check for each peripheral

libraries/ms-common/src/arm/uart.c
```diff
@@ -78,6 +78,7 @@ StatusCode uart_init(UartPort uart, UartSettings *settings) {
   // Init with disabled TX interrupts otherwise we will get TX
   // buffer empty continuously when there is no data to send
   USART_ITConfig(s_port[uart].base, USART_IT_TXE, DISABLE);
+  USART_ITConfig(s_port[uart].base, USART_IT_RXNE, ENABLE);
```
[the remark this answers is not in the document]
et312: Which specific project breaks from this line? I tried scons sim on leds and the log_debug statements display blink correctly.

projects/uart_cli/src/gpio_cli.c
```diff
@@ -0,0 +1,225 @@
...
+    printf("Invalid state - must be high or low\n\r");
+    return false;
+  } else {
+    return true;
+  }
+}
+
+int valid_pin_mode(char *pin_mode) {
+  const char *pin_modes[NUM_GPIO_MODES] = {
+    "analog",           "input_floating",   "input_pull_down", "input_pull_up", "output_open_drain",
```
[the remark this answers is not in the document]
et312: scons format gives this spacing - the build fails if I change it

mitchellostler (approved):

PR #283, opened 2024-05-22
Current GUI but with outdated dbc file reference

EricGaoo:


et312:
Lint and format passed, need to fix linker error

mitchellostler (approved):

PR #290, opened 2024-07-29
Completed tasks

HermanG05:
Completed tasks from FW 101 to 103.


projects/task2/src/queues1.c
```diff
@@ -0,0 +1,77 @@
...
+    StatusCode ret;
+    int to_send = 0;
+    while (true) {
+        ret = queue_send(&s_queue1, &s_list[to_send], 0);
+        delay_ms(100);
+        if (ret != STATUS_CODE_OK) {
+            LOG_DEBUG("Write to queue failed!");
+        }
+        to_send++;
+        if (to_send >= QUEUE_LEN) {
```
Akashem06, line 42: The queue sending is correct but this check needs to be fixed. You shouldn't break out of a Tasks while(true) loop

Having all of the logic under a if (to_send < QUEUE_LEN) would work

projects/gpio/src/main.c
```diff
@@ -0,0 +1,91 @@
...
+
+
+    gpio_toggle_state(&addr1);
+    delay_ms(1000);
+    gpio_toggle_state(&addr2);
+    delay_ms(1000);
+    gpio_toggle_state(&addr3);
+    delay_ms(1000);
+    gpio_toggle_state(&addr4);
+    delay_ms(1000);
```
Akashem06, line 74: Notify check is good!

But all of the gpio_toggle logic should be placed inside of the if(check). Check will return true only if the interrupt is triggered, which is when the gpio's should be toggled. There is also no reason for a 1sec delay after each toggle

Also the buttonAddr cannot and does not need to be toggled, its an input pin.

Akashem06 (changes requested):

PR #301, opened 2024-09-18
103 Parts 1&2 Submission - Anmeet Sekhon

AnmeetS:


projects/fw_103/src/ads1115.c
```diff
@@ -0,0 +1,74 @@
...
+
+StatusCode ads1115_select_channel(ADS1115_Config *config, ADS1115_Channel channel) {
+  if (config == NULL) {
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  uint16_t cmd;
+
+  // Write Config register
+  cmd = 0x8483; //I'm assuming the same configruation as earlier should be written here. I feel like it might be wrong
```
Bafran, line 48: For this task this is correct. If you were writing this function for production you'd need to use the `channel` parameter to actually select the correct channel. This information is available on the datasheet where you can select a different input for the ADC. In 103 we're just sticking with channel 0

projects/fw_103/src/main.c
```diff
@@ -0,0 +1,66 @@
...
+GpioAddress led_addr = {
+  .port = GPIO_PORT_B,
+  .pin = 3,
+};
+
+TASK(led_gpio_task, TASK_STACK_512) {
+  gpio_init_pin(&led_addr, GPIO_OUTPUT_PUSH_PULL, GPIO_STATE_LOW);
+  float reading;
+
+  //Ngl, i'm not fully sure why this is being done in the same task as the led thing from before but this seems to be what the instructions are saying.
```
Bafran, line 32: This is moreso just as an exercise. You're right it doesn't make a ton of sense to put them together. I may update the onboarding to make them separate.

Bafran (commented):
Looks good!

PR #302, opened 2024-09-18
Fwxv 285 Fix I2C x86

warren-xu:
Added comments with suggested changes to avoid seg faults when declaring x86 tasks with depths > 128.


Bafran (commented):
Hey, let me know if this should go in otherwise I'll close it out soon! I remember discussing this a while ago but no longer remember the state

PR #306, opened 2024-09-21
Bootloader

Akashem06:


projects/bootloader/scripts/main.py
Bafran, line 1: remove this?

projects/bootloader/scripts/test_can_datagram.py
```diff
@@ -0,0 +1,133 @@
...
+TEST_DATAGRAM_TYPE_ID = 30  # This is also the start ID
+TEST_NODES = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+TEST_NODES_RAW1 = 0b11111111
+TEST_NODES_RAW2 = 0b00000011
+TEST_DATA = bytearray([3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5, 8, 9,
+                       7, 9, 3, 2, 3, 8, 4, 6, 2, 6, 4, 3, 3])
+
+
+class TestCanDatagram(unittest.TestCase):
+    '''Test Can Datagram functions'''
```
Bafran, line 21: just curious, I see that these are copied from FWXIV, is there a reason you changed all the docstrings to single quotes?

platform/linker_scripts/bootloader/memory_sections.ld
```diff
@@ -0,0 +1,29 @@
+ENTRY(Reset_Handler)
+
+_bootloader_size = 8K;
```
Bafran, line 3: did you come up with this number after compilation?

Bafran (approved):
I'm gonna approve this since you said it works lol. It's a bit too large and I don't have enough context to properly review. Did you say there's a confluence doc though? Ryan's old bootloader page doesn't 1:1 match this and I think it would be better to make things a little more clear here

projects/bootloader/scripts/test_can_datagram.py
Akashem06: No reason I like single quotes more

platform/linker_scripts/bootloader/memory_sections.ld
Akashem06: Yes, I initially wanted 4k but it was too small, 8k is good for now

PR #310, opened 2024-09-21
Fw 999 fw 103 hw kien nguyen

k13nNg:
I have finished FW 103, can you help me look at it?

### DIFF
--- a/projects/fw_103/src/ads1115.c
+++ b/projects/fw_103/src/ads1115.c
@@ -14,27 +14,23 @@ StatusCode ads1115_init(ADS1115_Config *config, ADS1115_Address i2c_addr, GpioAd
   uint16_t cmd;
 
   // Write Config register
-  cmd = 0x8084; /* TODO: make setting this to continuous conversion an onboarding task */
+  /* TODO: fill out this value */
+  cmd = 0x0000;
   i2c_write_reg(config->i2c_port, i2c_addr, ADS1115_REG_CONFIG, (uint8_t *)(&cmd), 2);
 
-  /* TODO make this onboarding task*/
+  /* TODO (optional) */
   // Set low thresh to zero
   cmd = 0x0000;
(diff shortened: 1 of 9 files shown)


projects/fw_103/src/ads1115.c
```diff
@@ -0,0 +1,71 @@
...
+
+StatusCode ads1115_read_raw(ADS1115_Config *config, ADS1115_Channel channel, int16_t *reading) {
+  /* TODO: complete function */
+  i2c_read_reg(config->i2c_port, config->i2c_addr, ADS1115_REG_CONVERSION, (uint8_t *) reading, 8);
+  return STATUS_CODE_OK;
+}
+
+StatusCode ads1115_read_converted(ADS1115_Config *config, ADS1115_Channel channel, float *reading) {
+  /* TODO: complete function */
+  uint16_t reader ;
```
Akashem06, line 60: I believe if you pass in an int16_t instead of uint16_t, the value will be signed when returned. That way you won't need your check if (reader <= 32768). Just a simpler approach, don't need changes here. Everything else is good!

projects/fw_103/src/ads1115.c
```diff
@@ -0,0 +1,71 @@
...
+    return status_code(STATUS_CODE_INVALID_ARGS);
+  }
+
+  config->i2c_addr = i2c_addr;
+
+  uint16_t cmd;
+
+  // Write Config register
+  /* TODO: fill out this value */
+  cmd = 0x0011;
```
Akashem06, line 20: Incorrect CMD bitset. All bits in the config register should be default (Except for the conversion mode, which you will configure as continuous mode). The default bitset is 0000 0101 1000 0011, as per the datasheet.

Akashem06 (changes requested):
Review comments. Ask questions in your onboarding thread!

PR #316, opened 2024-09-28
FWXV-279: XBee Telemetry Branch

et312:


libraries/codegen/boards/bms_carrier.yaml
```diff
@@ -18,6 +18,8 @@
           watchdog: 3
         power_distribution:
           watchdog: 5
+        telemetry:
```
Bafran, line 21: Do we need to update the yamls? The task in main.c pops from a queue and doesn't do any checks on the messages. I might be forgetting but does that queue only contain messages that have the correct target board? or all messages?

If it's all messages then do we need to do this?

PR #324, opened 2024-10-01
Fw 999 fw 103 hw sophie

happyface3650:
Completed Firmware onboarding 101 to 103.


Bafran:

PR #338, opened 2024-10-30
SD Card Logger + Library

ikesicle:
`sd_binary.c` is the file which defines general-purpose methods for reading and writing to the SD card.
`main.c` is the smoke test file (need help writing this probably).


projects/fw_103/README.md
Bafran, line 1: I think these should all be removed

projects/sd_card_logger/src/sd_binary.c
```diff
@@ -0,0 +1,515 @@
+/**
+ *******************************************************************************
+ * @attention
+ *
+ * COPYRIGHT(c) 2016 STMicroelectronics
```
Bafran, line 5: Is this using something from the HAL?

projects/sd_card_logger/src/sd_binary.c
```diff
@@ -0,0 +1,515 @@
...
+static StatusCode prv_sd_get_next_write_data_response(SpiPort spi, uint8_t *response) {
+  volatile uint8_t readvalue;
+  uint16_t timeout = 8;
+  StatusCode s;
+
+  do {
+    s = prv_read_byte(spi, &readvalue);
+    if (s != STATUS_CODE_OK) return s;
+    if (--timeout)
+      return status_msg(
```
Bafran, line 218: Sorry I might not be familiar with the source of some of this code - where is this status_msg type coming from? looks like this function is still supposed to be returning our StatusCode

projects/sd_card_logger/src/sd_binary.c
```diff
@@ -0,0 +1,515 @@
...
+  // Split the cmd parameter into 8 byte ints
+  frame[0] = (cmd | 0x40);
+  frame[1] = (uint8_t)(arg >> 24);
+  frame[2] = (uint8_t)(arg >> 16);
+  frame[3] = (uint8_t)(arg >> 8);
+  frame[4] = (uint8_t)(arg);
+  frame[5] = (uint8_t)0x95;  // CRC for a CMD0. We can hardcode this since the CRC is ignored from
+                             // then on anyways.
+
+  spi_cs_set_state(spi, GPIO_STATE_LOW);
```
Bafran, line 260: Is this not taken care of within spi_tx? If there is a reason this is here then leave a comment about it

projects/sd_card_logger/src/sd_binary.c
```diff
@@ -0,0 +1,515 @@
...
+StatusCode sd_card_init(SpiPort spi) {
+  volatile StatusCode last_status;
+  volatile SdResponse last_response;
+  volatile uint16_t retry_counter = SD_NUM_RETRIES;
+
+  sd_port[spi] = SpiPortSdConfig_default;
+
+  // Send CMD0 (SD_CMD_GO_IDLE_STATE) to put SD in SPI mode and
+  // wait for In Idle State Response (R1 Format) equal to 0x01
+  do {
```
Bafran, line 314: Kind of an interesting setup you've got with the do while loops, did you reference this from somewhere or is it just a fitting solution given that you need to send once and continue on some condition?

smoke/sd_card_smoke/src/main.c
Bafran, line 1: This is super cool! Looks like a fairly simple interface and great job getting this to work I know it's a tough driver. Only note for this is maybe remove the mallocs and statically allocate so someone could copy paste your code into their project to test.

libraries/ms-common/src/arm/spi.c
```diff
@@ -153,11 +153,22 @@ StatusCode spi_exchange(SpiPort spi, uint8_t *tx_data, size_t tx_len, uint8_t *r
       rx_data[i - tx_len] = data;
     }
   }
+  mutex_unlock(&s_port[spi].spi_buf.mutex);
+  return STATUS_CODE_OK;
+}
 
-  // set spi CS state HIGH
-  gpio_set_state(&s_port[spi].cs, GPIO_STATE_HIGH);
+StatusCode spi_cs_set_state(SpiPort spi, GpioState state) {
```
Bafran, line 160: Oh I see I left a comment below about this, I think a comment would still help though

Bafran (commented):
This is really cool! I think it just requires a bit of cleanup since there's some random files, but if this works and is tested the general structure seems solid.

Bafran:
One more thing, for this project more than others please fill out the Readme with all the resources you used to write the driver. This one has some quirks so if there's no documentation it'll get out of date pretty quickly

projects/sd_card_logger/src/sd_binary.c
ikesicle: Not sure what exactly the HAL is, could you clarify? Also, that line above seems to be added by scons format.

projects/sd_card_logger/src/sd_binary.c
ikesicle: Not sure either. The old SD card program which this was based off of seemed to use it, and it expanded into a macro which returned a Status Code as normal. Maybe it returns a status code and also logs a message at the same time?

projects/sd_card_logger/src/sd_binary.c
ikesicle: Don't actually think there's a reason for that to be there, I remember spi_tx does pull down. I think I could remove it.

projects/sd_card_logger/src/sd_binary.c
ikesicle: It just seemed fitting since every step requires an initial try of the process, so I thought do-while would work for that

PR #343, opened 2024-11-05
Steven telemetry CAN autogen

StevenTB1:
Changes:
- Added python_autogen.py.jinja
- bms_carrier_tx_all.c is useless and can be removed
- Added message_dict in generator.py

Questions:
- How to test the script generated?

### DIFF
--- a/libraries/codegen/bms_carrier_tx_all.c
+++ b/libraries/codegen/bms_carrier_tx_all.c
@@ -1,0 +1,60 @@
+#include <stdint.h>
+
+#include "can_board_ids.h"
+#include "can_codegen.h"
+
+static CanMessage s_msg = { 
+    .type = CAN_MSG_TYPE_DATA,
+};
+static void prv_tx_can_message(CanMessageId id, uint8_t num_bytes, uint64_t data) {
+    s_msg.id.raw = id,
+    s_msg.dlc = num_bytes;
+    s_msg.data = data;
+    s_msg.extended = (s_msg.id.msg_id >= CAN_MSG_MAX_STD_IDS);
+    can_transmit(&s_msg);
+}
+
+void can_tx_all() {
+    prv_tx_can_message(
+        SYSTEM_CAN_MESSAGE_BMS_CARRIER_BATTERY_STATUS,7,
+        (uint64_t) g_tx_struct.battery_status_fault << 0 |
+        (uint64_t) g_tx_struct.battery_status_fault_val << 16 |
+        (uint64_t) g_tx_struct.battery_status_aux_batt_v << 32 |
+        (uint64_t) g_tx_struct.battery_status_afe_status << 48);
+    prv_tx_can_message(
+        SYSTEM_CAN_MESSAGE_BMS_CARRIER_BATTERY_INFO,6,
+        (uint64_t) g_tx_struct.battery_info_fan1 << 0 |
+        (uint64_t) g_tx_struct.battery_info_fan2 << 8 |
+        (uint64_t) g_tx_struct.battery_info_max_cell_v << 16 |
+        (uint64_t) g_tx_struct.battery_info_min_cell_v << 32);
+    prv_tx_can_message(
+        SYSTEM_CAN_MESSAGE_BMS_CARRIER_BATTERY_VT,8,
+        (uint64_t) g_tx_struct.battery_vt_voltage << 0 |
+        (uint64_t) g_tx_struct.battery_vt_current << 16 |
+        (uint64_t) g_tx_struct.battery_vt_temperature << 32 |
+        (uint64_t) g_tx_struct.battery_vt_batt_perc << 48);
+    prv_tx_can_message(
+        SYSTEM_CAN_MESSAGE_BMS_CARRIER_BATTERY_RELAY_INFO,1,
+        (uint64_t) g_tx_struct.battery_relay_info_state << 0);
+    prv_tx_can_message(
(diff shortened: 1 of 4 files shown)


libraries/codegen/bms_carrier_tx_all.c
Bafran, line 1: This file is autogenerated right? It shouldn't be in our repo, there should be a gitignore to ignore this from a commit so it's odd that it showed up. Might just be in the wrong directory

Bafran, line 1: Hey, saw you working on this yesterday! Gonna close this out but it looks like you're missing the 103 files, 101 and 102 are here. Let me know if you need help with anything!

libraries/codegen/generator.py
```diff
@@ -94,8 +94,31 @@ def get_data():
...
+                print(f"Message Name: {message['name']}")
+                print(f"  ID: {message['id']}")
+                print(f"  Critical: {message['critical']}")
+                print(f"  Sender: {message['sender']}")
+                print(f"  Receiver: {message['receiver']}")
+                print("  Signals:")
+                for signal in message['signals']:
+                    print(f"   - Name: {signal['name']}, Start Bit: {signal['start_bit']}, Length: {signal['length']}")
+
+    return {"Boards": boards, "Messages": messages, "Messages_dict": messages_dict}
```
Bafran, line 121: Nice you're right this should speed things up, not a big deal for our regular autogen but definitely useful for telemetry.

libraries/codegen/templates/_tx_all.c.jinja
```diff
@@ -26,5 +26,4 @@ void can_tx_all() {
         (uint64_t) g_tx_struct.{{message.name}}_{{signal.name}} << {{signal.start_bit}}{{ " |" if not loop.last }}
     {%- endfor -%}
     );
-{%- endfor %}
-}
+{%  }
```
Bafran, line 29: Why the edit in the .c file?

libraries/codegen/templates/python_autogen.py.jinja
```diff
@@ -0,0 +1,50 @@
+{% set bitstream = data["bitstream"] -%}
+{% set messages_dict = data["Messages_dict"] -%}
+
+def decode_packet(messages_dict, bitstream):
```
Bafran, line 4: So here does the user pass in the messages dict themselves?

You're right that maybe a quick sample app/unit test would be good and answer this question implicitly.

libraries/codegen/templates/python_autogen.py.jinja
```diff
@@ -0,0 +1,50 @@
+{% set bitstream = data["bitstream"] -%}
+{% set messages_dict = data["Messages_dict"] -%}
+
+def decode_packet(messages_dict, bitstream):
+    # Clean the bitstream by removing spaces
+    clean_bitstream = bitstream.replace(" ", "")
```
Bafran, line 6: So, your function received a CAN frame right? Why are there spaces in the bitstream?

Bafran (commented):
Awesome thanks for doing this I think it generally looks pretty good. I think a quick unit test would make the usage a little clearer, plus we should think about integration a little bit before merging this. My only concern is the format of the input data, so we should see what the shape of our CAN frame is once we received it from the Xbee and have it in memory (e.g bitstream in an array, array of ints, etc.)

For now just add in a unit test (does not need to be rigorous at all) which should primarily serve as a usage example

libraries/codegen/templates/python_autogen.py.jinja
StevenTB1: in case something like 0000 0001 occurs, just removing the blank space in between.

libraries/codegen/bms_carrier_tx_all.c
StevenTB1: I think this file is autogenerated when im playing around with the generator.py to see how it generates files. If I accidentally deleted something it might unintentional.

libraries/codegen/templates/_tx_all.c.jinja
StevenTB1: Think when looking into the CAN message in C side and accidently changed something. Sorry about that.

libraries/codegen/templates/python_autogen.py.jinja
StevenTB1: I was also having a bit confusion in relation to this one. The jinja defines a python function that passes these two in as parameter, but currently they are located in generator.py. (data from get_data() that gets all the yaml files)

I am planning on after generating the python script, I just run `python3 libraries/codegen/python_autogen.py` with some arbitrary defined CAN messages in the main function. Something similar to this:

# Define message values
    # source_id = 1                 # source_id (3 + 4 bits)
    # type_field = 0                # type (1 bit)
    # msg_id = 3                    # msg_id (24 bits)
    # dlc = 5 Bytes                 # (40 bits)
    # throttle_output_value = 30    # throttle_value (32 bits)
    # brake_output_value = 31       # brake_value (8 bits)
    # print(decode_packet(data["Messages_dict"], "0000 0001 0000 0000 0000 0000 0000 0011 0000 0000 0000 0000 0000 0101 0000 0000 0000 0000 0000 0000 0001 1110 0001 1111"))

If eventually this python_autogen.py file will be in libraries/codegen, I'm thinking about just doing `import data from generator.py` to get the data array. Maybe I will look into this solution on Saturday in the bay.

PR #350, opened 2024-12-01
Finished Scons Debug/Release Configuration Task, added debug and release flags

TamirPol:


Akashem06 (approved):
Looks good to me but I'm going to hold off on merging until I test the release mode on every project. Only wanna do this because the optimizations can cause some stack overflows if we are not careful

